PR #187, opened 2020-10-08
Sampler and Workflow Refactor

nmerket:
Fixes #147.

## Pull Request Description

- Sampler class refactor. Simplifies things quite a it. 

## Checklist

Not all may apply

- [x] Code changes (must work)
- [x] Tests exercising your feature/bug fix (check coverage report on CircleCI build -> Artifacts)
- [x] All other unit tests passing
- [x] Update validation for project config yaml file changes
- [x] Update existing documentation
- [x] Run a small batch run to make sure it all works (local is fine, unless an Eagle specific feature)
- [x] Add to the changelog_dev.rst file and propose migration text in the pull request

PR #205, opened 2021-01-13
Fix: racially insensitive language

### LINKED ISSUE #164: Remove racially insensitive language
The big one is the fact that we're mirroring the [AWS EMR API](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/emr.html#EMR.Client.run_job_flow) by using the terms "master" and "slave" [in our input file](https://github.com/NREL/buildstockbatch/blob/4ab5974c8588642257ed3460f7d71e6f02b5aa70/buildstockbatch/schemas/v0.2.yaml#L38-L40). I'm open to suggestions on what the updated terms should be. Maybe we should look at other projects that have a similar paradigm to see what terminology they use so we can be consistent. 

Next, we should change the name of the `master` branch to something else. [Github as a whole is moving this direction](https://thenextweb.com/dd/2020/06/15/github-plans-to-replace-racially-insensitive-terms-like-master-and-whitelist/). This *should* be pretty easy, but could be disruptive to developers so will need to be coordinated. We have several repos, so maybe this change should be done in conjunction with others.

aspeake:
Fixes #164 

## Pull Request Description
Removes racially insensitive language for EMR instance types, updating the following arguments:
- `master_instance_type` ==> `manager_instance_type`
- `slave_instance_type` ==> `worker_instance_type`
- `slave_instance_count` ==> `worker_instance_count`

Also, now fixes the broken AWS workflow.

## Checklist

Not all may apply


buildstockbatch/schemas/v0.1.yaml
```diff
@@ -33,9 +33,9 @@ aws-job-environment:
   memory: int(min=1024, required=False)
 
 aws-emr-spec:
-  master_instance_type: str(required=False)
-  slave_instance_type: str(required=False)
-  slave_instance_count: int(min=1, required=False)
+  manager_instance_type: str(required=False)
+  worker_instance_type: str(required=False)
+  worker_instance_count: int(min=1, required=False)
```
nmerket: We shouldn't be changing old schemas as the only function they serve now is to keep a record of previous input formats.

buildstockbatch/schemas/v0.2.yaml
```diff
@@ -35,9 +35,9 @@ aws-job-environment:
   memory: int(min=1024, required=False)
 
 aws-emr-spec:
-  master_instance_type: str(required=False)
-  slave_instance_type: str(required=False)
-  slave_instance_count: int(min=1, required=False)
+  manager_instance_type: str(required=False)
+  worker_instance_type: str(required=False)
+  worker_instance_count: int(min=1, required=False)
```
nmerket: Same here. Since we're making breaking changes to the input format, we should make a v0.3 schema and upgrade to that. I actually have a v0.3 schema going on in #187. Maybe we should wait for that to merge (soon) and make these changes there.

nmerket (commented):
Overall looks pretty good. We might want to wait a bit to merge this because of input schema changes. I would feel more comfortable if we could test this somehow as well.

buildstockbatch/schemas/v0.1.yaml
aspeake: Makes sense, addressed in https://github.com/NREL/buildstockbatch/pull/205/commits/517049db72bbd2f5ab210d85db08a739e421dd86

buildstockbatch/schemas/v0.2.yaml
aspeake: Addressed in https://github.com/NREL/buildstockbatch/pull/205/commits/517049db72bbd2f5ab210d85db08a739e421dd86. That sounds good, we can hold off for the next schema to get merged.

nmerket (approved):

PR #212, opened 2021-03-15
memory_issue_fix

### LINKED ISSUE #210: Postprocessing MAX_PARQUET_MEMORY too high?
Hello there,

I'm encountering nasty Dask errors with the current `MAX_PARQUET_MEMORY` setting when postprocessing results. When the codes reaches the parquet timeseries step, several warnings are printed:
```
distributed.worker - WARNING - Memory use is high but worker has no data to store to disk.  Perhaps some other process is leaking memory?  Process memory: 6.19 GB -- Worker memory limit: 6.33 GB  
```
Then eventually the script fails with the following error:
```
distributed.scheduler.KilledWorker: ("('from-delayed-read_and_concat_enduse_timeseries_parquet-set_index-9bdd5f8bca01636a0a0ff8d724e3093a-index-5ad59557d04-1781', 25)", <Worker 'tcp://127.0.0.1:34983', name: 4, memory: 0, processing: 4>)
```

Looking at the code below, I wonder if `MAX_PARQUET_MEMORY` should be determined as a function of how many CPUs dask will utilize and the total available RAM? 

https://github.com/NREL/buildstockbatch/blob/94597284e6a61521bd94358f015501b701f3eb1c/buildstockbatch/postprocessing.py#L335-L338

In my case, reducing this parameter from `1E9` to `2E8` avoided any further failures.

rajeee:
Fixes #210 .

## Pull Request Description

The v0.19 postprocessing has been failing on eagle with issue #210 when trying to postprocess the recent restock runs with schedules. After series of debugging session, it seems like dask and eagle's nodes/process just don't work well together.
This pull request does a few things:
1. Reduces the number of process and threads on each nodes to just 1. This was necessary to avoid memory issues. Probably Dask's multiprocess/multithread memory management is causing problems.
2. Instead of using a giant dask datafarame created from list of delayed objects, to write all of the parquet files using single command, it uses a simple list of delayed function that reads a group of parquet files, combines them and writes them. This was necessary to avoid communication/memory issue we were having while using the giant dask dataframe.
3. Exposes the postprocessing node memory size in the yaml. 
4. Adds a "keep_intermediate_files" boolean option in the yaml. Intermediate files are very useful during debugging and sometimes also useful in use cases where individual simulation output timeseries is required.

Oh, I also did some benchmarking for the optimal size of the aggregated parquet files. Looks like both 1GB and 4GB (the new default) is good.
image
Query1: SELECT "build_existing_model.state" as state, sum("total_site_electricity_kwh") as total_energy FROM "test_runs"."<test_table>_timeseries" join  "<test_table>" on "<test_table>_timeseries"."building_id" = "<test_table>_baseline"."building_id"  group by 1 order by 1;
Query2: SELECT count(*) from "test_runs"."<test_table>_timeseries";

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/eagle.py
+++ b/buildstockbatch/eagle.py
@@ -494,7 +494,8 @@ class EagleBatch(BuildStockBatchBase):
         # Configuration values
         account = self.cfg['eagle']['account']
         walltime = self.cfg['eagle'].get('postprocessing', {}).get('time', '1:30:00')
-
+        memory = self.cfg['eagle'].get('postprocessing', {}).get('memory', 85248)
+        print(f"Submitting job to {memory}MB memory nodes.")
         # Throw an error if the files already exist.
         if not upload_only:
             for subdir in ('parquet', 'results_csvs'):
@@ -530,7 +531,7 @@ class EagleBatch(BuildStockBatchBase):
             '--output=postprocessing.out',
             '--nodes=1',
             ':',
-            '--mem=180000',
+            '--mem={}'.format(memory),
             '--output=dask_workers.out',
             '--nodes={}'.format(self.cfg['eagle'].get('postprocessing', {}).get('n_workers', 2)),
             eagle_post_sh
@@ -599,7 +600,6 @@ def user_cli(argv=sys.argv[1:]):
     '''
     This is the user entry point for running buildstockbatch on Eagle
     '''
-
     # set up logging, currently based on within-this-file hard-coded config
     logging.config.dictConfig(logging_config)
 
--- a/buildstockbatch/eagle.py
+++ b/buildstockbatch/eagle.py
@@ -519,6 +519,7 @@ class EagleBatch(BuildStockBatchBase):
         env['MY_CONDA_ENV'] = os.environ['CONDA_PREFIX']
         env['OUT_DIR'] = self.output_dir
         env['UPLOADONLY'] = str(upload_only)
+        env['MEMORY'] = str(memory)
         here = os.path.dirname(os.path.abspath(__file__))
         eagle_post_sh = os.path.join(here, 'eagle_postprocessing.sh')
 
@@ -526,7 +527,7 @@ class EagleBatch(BuildStockBatchBase):
             'sbatch',
             '--account={}'.format(account),
             '--time={}'.format(walltime),
-            '--export=PROJECTFILE,MY_CONDA_ENV,OUT_DIR,UPLOADONLY',
+            '--export=PROJECTFILE,MY_CONDA_ENV,OUT_DIR,UPLOADONLY,MEMORY',
             '--job-name=bstkpost',
             '--output=postprocessing.out',
             '--nodes=1',
--- a/buildstockbatch/eagle.py
+++ b/buildstockbatch/eagle.py
@@ -494,7 +494,7 @@ class EagleBatch(BuildStockBatchBase):
         # Configuration values
         account = self.cfg['eagle']['account']
         walltime = self.cfg['eagle'].get('postprocessing', {}).get('time', '1:30:00')
-        memory = self.cfg['eagle'].get('postprocessing', {}).get('memory', 85248)
+        memory = self.cfg['eagle'].get('postprocessing', {}).get('node_memory_mb', 85248)
         print(f"Submitting job to {memory}MB memory nodes.")
         # Throw an error if the files already exist.
         if not upload_only:
--- a/buildstockbatch/postprocessing.py
+++ b/buildstockbatch/postprocessing.py
@@ -9,7 +9,6 @@ A module containing utility functions for postprocessing
 :copyright: (c) 2018 by The Alliance for Sustainable Energy
 :license: BSD-3
 """
-import traceback
 import boto3
 import dask.bag as db
 from dask.distributed import performance_report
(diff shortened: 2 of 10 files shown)


buildstockbatch/eagle.py
```diff
@@ -494,7 +494,8 @@ def queue_post_processing(self, after_jobids=[], upload_only=False, hipri=False)
         # Configuration values
         account = self.cfg['eagle']['account']
         walltime = self.cfg['eagle'].get('postprocessing', {}).get('time', '1:30:00')
-
+        memory = self.cfg['eagle'].get('postprocessing', {}).get('memory', 85248)
```
nmerket: Is this right? I don't see a `memory` key in the postprocessing spec in the schema. Is this supposed to be

```suggestion
        memory = self.cfg['eagle'].get('postprocessing', {}).get('node_memory_mb', 85248)
```

buildstockbatch/postprocessing.py
```diff
@@ -218,17 +218,28 @@ def read_results_json(fs, filename):
...
+    try:
+        with fs.open(filename, 'rb') as f:
+            df = pd.read_parquet(f, engine='pyarrow')
+        building_id = int(re.search(r'bldg(\d+).parquet', filename).group(1))
+        df['building_id'] = building_id
+        for col in set(all_cols).difference(df.columns.values):
+            df[col] = np.nan
+        return df[all_cols]
+    except Exception:
+        return pd.DataFrame(columns=[all_cols]+['building_id'])
```
nmerket: This is a pretty broad `except` block. It looks like it just returns an empty dataframe with the appropriate columns if anything goes wrong. Intentionally creating these kinds of silent errors is troubling because it means that there could be major problems with reading some of the parquet files and the user will not know and think everything is fine.

buildstockbatch/postprocessing.py
```diff
@@ -330,32 +341,39 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
...
+            total_mem = mean_mem * len(ts_filenames) / 1e6  # total_mem in MB
 
             # Determine how many files should be in each partition and group the files
-            npartitions = math.ceil(total_mem / MAX_PARQUET_MEMORY)  # 1 GB per partition
+            parquet_memory = int(cfg['eagle'].get('postprocessing', {}).get('parquet_memory_mb', MAX_PARQUET_MEMORY))
+            logger.info(f"Max parquet memory: {parquet_memory} MB")
+            npartitions = math.ceil(total_mem / parquet_memory)
             npartitions = min(len(ts_filenames), npartitions)  # cannot have less than one file per partition
             ts_files_in_each_partition = np.array_split(ts_filenames, npartitions)
+            N = len(ts_files_in_each_partition[0])
```
nmerket: Minor thing, but usually a variable written in `ALL_CAPS` is understood to be a constant in python. `CamelCase` is a class. This looks to be a regular variable which are usually `lower_snake_case`. It's not clear from the variable name what this is. Maybe a more descriptive name would be helpful.

nmerket (changes requested):
Looks pretty good. A few comments below. Could use some documentation updates and something in the changelog.

This reminds me I need to get things set up to have multiple versions of the docs published at the same time.

buildstockbatch/eagle.py
rajeee: Good catch! Thanks. I had named it 'memory' at first but changed it to node_memory_mb to distinguish it from 'parquet_memory_mb'. Forgot to change it here, and it silently passed my test runs cause it just used the default.

buildstockbatch/postprocessing.py
rajeee: Yeah, agreed it could mask errors. But I think a few times there was an issue reading a handful of parquet files that resulted in crashing everything, so I thought it might be a good idea to assemble whatever can be assembled. Do you think it would be okay to have this try-except if we only wrap the read_parquet call?

buildstockbatch/postprocessing.py
rajeee: point taken :). Thanks. I have a propensity to use big N for counts of any sort. I will change the name.

buildstockbatch/postprocessing.py
nmerket: I figured that's why you had this here. I don't know if it functionally matters much to only wrap the read_parquet call rather than do this. My primary concern is that this will silently omit data the user is expecting. That could mess up results. Right now at least the user knows something went wrong because they don't get their results. Could we put some sort of error message in the postprocessing (or other relevant) log file saying, "building_id = x, upgrade = y failed to load parquet"? They still might miss that, but it would be documented somewhere. It would be better than what we have now, which is an obtuse error with very little to go on to find the problem.

rajeee: @nmerket I started working on the solution you proposed of logging the buildings we failed to read the timeseries, but then realized we would have to do it at multiple places. If there is any corrupted parquet it would fail here first: https://github.com/NREL/buildstockbatch/blob/1cd0349759324e93348ce91d1c3ee59e683af8f3/buildstockbatch/postprocessing.py#L281

And, the corrupted parquet (or not being able to read from disk) is something I have seen only a couple of times when the eagle was having an issue, so it may not be worthwhile to account for that. Even if we need to handle that, maybe not in this PR. So, I have removed all such handling.

nmerket (approved):
Looks good. Thanks for making these changes!

PR #228, opened 2021-05-03
Keep individual timeseries files

### LINKED ISSUE #182: Switch for automatic deletion of time series parquet files
This [bit of code in the postprocessing](https://github.com/NREL/buildstockbatch/blob/develop/buildstockbatch/postprocessing.py#L365-L368) doesn't have a switch on it and we really really truly need these parquet files for vizstock in particular. The aggregated timeseries files that can be produced from them are useless to us at this point for production runs of ComStock, and soon ResStock.

A flag to disable this behavior.

**Describe alternatives you've considered**
Making the flag default to not clean up, but I think that's not really reasonable...

**Additional context**
I've spend over 160 hours 'fixing' this for the V1 ComStock run by re-processing over a PB of double-zipped data. I look forward to never doing this ever again.

nmerket:
Fixes #182.

## Pull Request Description

Moves the `eagle.postprocessing.keep_intermediate_files` to `postprocessing.keep_individual_timeseries` and changes behavior to keep only the timeseries parquet files. Also, removes the deprecated `aggregate_timeseries` key as that aggregation always happens.

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/postprocessing.py
+++ b/buildstockbatch/postprocessing.py
@@ -365,15 +365,16 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
             logger.info(f"Finished combining and saving timeseries for upgrade{upgrade_id}.")
 
 
-def remove_intermediate_files(fs, results_dir):
+def remove_intermediate_files(fs, results_dir, keep_individual_timeseries=False):
     # Remove aggregated files to save space
     sim_output_dir = f'{results_dir}/simulation_output'
-    ts_in_dir = f'{sim_output_dir}/timeseries'
     results_job_json_glob = f'{sim_output_dir}/results_job*.json.gz'
-    logger.info('Removing temporary files')
-    fs.rm(ts_in_dir, recursive=True)
+    logger.info('Removing results_job*.json.gz')
     for filename in fs.glob(results_job_json_glob):
         fs.rm(filename)
+    if not keep_individual_timeseries:
+        ts_in_dir = f'{sim_output_dir}/timeseries'
+        fs.rm(ts_in_dir, recursive=True)
 
 
 def upload_results(aws_conf, output_dir, results_dir):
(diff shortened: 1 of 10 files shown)


buildstockbatch/postprocessing.py
```diff
@@ -365,15 +365,16 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
...
-def remove_intermediate_files(fs, results_dir):
+def remove_intermediate_files(fs, results_dir, keep_individual_timeseries=False):
     # Remove aggregated files to save space
     sim_output_dir = f'{results_dir}/simulation_output'
-    ts_in_dir = f'{sim_output_dir}/timeseries'
     results_job_json_glob = f'{sim_output_dir}/results_job*.json.gz'
-    logger.info('Removing temporary files')
-    fs.rm(ts_in_dir, recursive=True)
+    logger.info('Removing results_job*.json.gz')
     for filename in fs.glob(results_job_json_glob):
```
rHorsey, line 373: @nmerket In the past this function would run even if there was an uncaught error in the metadata results aggregation, leading to the entire analysis having to be rerun. Is it possible still for this code to be reached if that dask cluster job doesn't complete successfully?

rHorsey (commented):
lgtm - one question which is outside the scope of this PR but still relevant to the key function at issue. Happy to approve @nmerket if this is a priority to get merged!

buildstockbatch/postprocessing.py
nmerket: It's possible. I didn't do anything in this PR to mitigate that possibility. In general, this _should_ run only after the results.csv aggregation is complete. Is there a way you could provide a minimum verifiable example of this issue?

nmerket: And the scope of this PR is to leave the timeseries files on disk, not the results jsons.

PR #235, opened 2021-05-27
Fix ResStock inverse weight

### LINKED ISSUE #234: ResStock - Inverse Sample Weight
## Issue description

The sample weight is inverse to what we would expect. In a recent run with BuildStock Batch 0.20 I used the following YAML inputs:

```
baseline:
  n_buildings_represented: 133172057 # ACS 5-yr 2016
sampler:
  type: residential_quota
  args:
    n_datapoints: 550000
```

I would expect the weight to be `33,172,057/550,000 = 242.13`. However, the weight ended up being `0.00413 ~ 550,000/133,172,057`.

## Bug Location

I believe this is the issue in the ResStock `workflow_generator`.

https://github.com/NREL/buildstockbatch/blob/7f7e628d9e5fa24cf24016effe6732d863028eb1/buildstockbatch/workflow_generator/residential.py#L256

afontani:
Fixes #234.

## Pull Request Description

This pull request fixes the issue that the ResStock weight is inverse to what it is supposed to be. The issue was identified in the workflow generator.

I tested the result locally with `buildstock_docker`. 

I would suggest this change be included in a 0.20.1 release as it breaks the residential validation workflows.

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/workflow_generator/residential.py
+++ b/buildstockbatch/workflow_generator/residential.py
@@ -253,7 +253,7 @@ class ResidentialDefaultWorkflowGenerator(WorkflowGeneratorBase):
         bld_exist_model_args = {
             'building_id': building_id,
             'workflow_json': 'measure-info.json',
-            'sample_weight': self.n_datapoints / self.cfg['baseline']['n_buildings_represented'],
+            'sample_weight': self.cfg['baseline']['n_buildings_represented'] / self.n_datapoints,
         }
         if 'measures_to_ignore' in workflow_args:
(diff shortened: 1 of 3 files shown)


nmerket (changes requested):
Oops. Glad this was an easy fix. Could you add as simple unit test [here](https://github.com/NREL/buildstockbatch/blob/bug/inverse_resstock_weight/buildstockbatch/workflow_generator/test_workflow_generator.py) to verify that this is working? There are some pretty examples setting a test up in there you could start from. The workflow generators are fairly self contained code so that should make testing a bit easier for this bug fix.

nmerket (approved):

PR #240, opened 2021-07-29
Update to OS v3.2.1

joseph-robertson:
## Pull Request Description

Companion PR: https://github.com/NREL/resstock/pull/661 (ServerDirectoryCleanup updates).

This PR is necessary to work with resstock develop.

By setting `skip_zip_results=True` in the workflow generator, each datapoint's `run` directory contains all simulation files and they aren't contained in a `data_point.zip`. Also, the size of `simulations_job0.tar.gz` actually decreases.

**BEFORE**
image

image

image

**AFTER**
image

image

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -38,8 +38,8 @@ logger = logging.getLogger(__name__)
 
 class BuildStockBatchBase(object):
 
-    DEFAULT_OS_VERSION = '2.9.1'
-    DEFAULT_OS_SHA = '3472e8b799'
+    DEFAULT_OS_VERSION = '3.2.1'
+    DEFAULT_OS_SHA = 'bdbdbc9da6'
     CONTAINER_RUNTIME = None
     LOGO = '''
      _ __         _     __,              _ __
--- a/buildstockbatch/workflow_generator/residential.py
+++ b/buildstockbatch/workflow_generator/residential.py
@@ -274,6 +274,9 @@ class ResidentialDefaultWorkflowGenerator(WorkflowGeneratorBase):
             'measure_paths': [
                 'measures'
             ],
+            'run_options': {
+                'skip_zip_results': True
+            }
         }
 
         osw['steps'].extend(workflow_args['measures'])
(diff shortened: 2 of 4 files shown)


nmerket:
This is good. With the data_point.zip now gone, I was worried this code would break:

https://github.com/NREL/buildstockbatch/blob/16cf0a867cec151f7021b2c7da06fa28ca403aec/buildstockbatch/base.py#L203-L210

It was looks in data_point.zip and removes things from the directory that are already in there. But if data_point.zip isn't there it skips that block, so I guess it's not hurting anything. I'm not sure what's going on in ComStock so maybe we leave it for now. 

In general, I'd say just go through that cleanup function to make sure it's still relevant with the organization of the files being written with your update.

joseph-robertson:
@nmerket I've updated the residential workflow generator to write argument values for ServerDirectoryCleanup. I'm not sure what else needs to be updated (e.g., pretty sure validation will now fail if `server_directory_cleanup` is found in a project yml file).

nmerket (commented):
@joseph-robertson This seems pretty straightforward and I like what you're doing with the server directory cleanup. You're right in that it will throw an error if that argument is there even though you're expecting it. That's easy enough to fix, though. Just add some yaml schema stuff here to tell it what to expect. It's even in the same file.

https://github.com/NREL/buildstockbatch/blob/81b43f0a20efe34ee599e1d0b30f2d32f124768e/buildstockbatch/workflow_generator/residential.py#L43-L54

I haven't tried running it, but it sounds like @munank has and it worked for him, so that's pretty good.

A few things to update:

- [x] The residential workflow generator schema as mentioned above.
- [x] Documentation for the residential workflow generator to document the schema changes.
- [x] Changelog entry

joseph-robertson:
@nmerket Anything else that needs to be done here?

asparke2:
@joseph-robertson I think this is fine as-is and we can fix up the ComStock server directory cleanup as necessary.  One thing to be aware of is that it won't work if `custom_gems = True`.  I don't think you use that for ResStock, so probably not a big deal.  [This branch](https://github.com/NREL/buildstockbatch/tree/feature/enable_os_3x_w_custom_gems) has additional changes to accommodate custom_gems with OS 3.x, but not going to worry about that until FY22.

nmerket (approved):

nmerket:
Looks good to me. You okay if I merge then?

joseph-robertson:
Let me get review on https://github.com/NREL/resstock/pull/661 and then I'll let you know.

PR #238, opened 2021-06-22
Enable ComStock OS 3.4.0 with custom gems

asparke2:
Fixes #237

## Pull Request Description

This PR makes several modifications to work for ComStock.

First, the default commercial workflow generator was changed to mimic the residential workflow generator, where a new `timeseries_csv_export` key was added to the workflow schema in order to trigger timeseries postprocessing.  The key/value pairs under this key set the arguments to the `TimeseriesCSVExport` measure in the workflow, whereas they were previously hard-coded in the workflow generator.

```
  args:
    reporting_measures:
       - measure_dir_name: qoi_report
    - measure_dir_name: simulation_settings_check
      arguments:
        run_sim_settings_checks: true
    timeseries_csv_export:
      reporting_frequency: Timestep
      inc_output_variables: false
```

Second, the CLI commands were changed to work with OS 3.X when `custom_gems=true` in the YML file.

Third, in order to enable the use of custom gems on local docker runs, at runtime, as the first step in building the Docker image, the target BuildStock repo is mounted, and the custom gems specified by that repo (via `MyBuildStockDir/resources/Gemfile`) are installed to a volume called `buildstockbatch_custom_gems` to replace the default gems (`/var/oscli/gems `) using bundler.  The `buildstockbatch_custom_gems` volume is not deleted after the run, because this way, if a developer wants to work on the custom gems, they can do so quickly without having to re-download all the gems locally for each test run. Not coincidentally, the `MyBuildStockDir/resources/Gemfile` in the BuildStock directory is also used in the creation of custom Singularity containers used by Eagle, and is envisioned to be the source of custom gem specifications to use with BuildStockBatch for AWS runs in a future PR.

## Checklist

Not all may apply

- [x] Code changes (must work)
- [x] Tests exercising your feature/bug fix (check coverage report on CircleCI build -> Artifacts)
- [x] All other unit tests passing
- [x] Update validation for project config yaml file changes
- [x] Update existing documentation
- [x] Run a small batch run to make sure it all works (local is fine, unless an Eagle specific feature)
- [ ] Add to the changelog_dev.rst file and propose migration text in the pull request

PR #245, opened 2021-09-07
Make it run for resstock.

rajeee:


nmerket:
@rajeee Is this still relevant?

rajeee:
I think so; unless there is another PR that makes the features/aws_with_custom_gems work for resstock. Either way, things could have changed since this PR so it definitely needs more testing before we can merge.

nmerket:
Yeah, it definitely looks outdated. It's pulling OS 2.9.1 and I think we're using 3.4 at this point. The ComStock team has added a few things in #238 regarding gems. I propose we merge that in first since it's closer to the finish line and then revisit if we bring this one up to date or just start a new branch with changes relevant to the current state of things.

afontani:
@rajeee : Can we close this?

asparke2:
please don't delete this branch yet, I would like to salvage the work

PR #257, opened 2022-01-20
Remove metafiles interfering with dask read

rajeee:
Fixes the failing test.

## Pull Request Description

Because the reference file used by the test had some parquet metadata files and the files generated don't, the comparison was failing. This fixes the issue by making the reference file similar to (i.e. no metadata files) to that generated by postprocessing.

The issue is popping up only now likely because dask.DataFrame.read_parquet started behaving differently when there is metadata file only recently?! 

## Checklist

Not all may apply


nmerket (approved):
Thanks for fixing this!

PR #258, opened 2022-01-21
Memory efficient results.csv creation

rajeee:
Potentially fixes #253.

## Pull Request Description

Instead of loading all the results_jobx.json.gz files at once to memory, use dask to load only the results_jobs for an upgrade at a time. This should prevent out of memory error when dealing with large run with many upgrades.

## Checklist

Not all may apply

- [x] Code changes (must work)
- [x] Tests exercising your feature/bug fix (check coverage report on CircleCI build -> Artifacts)
- [x] All other unit tests passing
- [ ] Update validation for project config yaml file changes
- [ ] Update existing documentation
- [x] Run a small batch run to make sure it all works (local is fine, unless an Eagle specific feature)
- [ ] Add to the changelog_dev.rst file and propose migration text in the pull request

Issue #223, opened 2021-04-26
Migrate CI to github actions


nmerket:
**Describe the solution you'd like**
We could run a (small) actual ResStock and/or ComStock simulation for our testing because it would be able to run and call docker instead of being in a docker container.

PR #262, opened 2022-01-29
Fix postprocessing failure on local docker

rajeee:
## Pull Request Description

The postprocessing was failing on docker because the last PR  (#258) was relying in jobx.json files to find out which simulation belonged to which job and finding the total number of upgrades. However, jobx.json files are not available while running locally using docker. This PR addresses that issue by not relying on jobx.json anymore.

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/postprocessing.py
+++ b/buildstockbatch/postprocessing.py
@@ -204,13 +204,13 @@ def get_cols(fs, filename):
     return schema.names
 
 
-def read_results_json(fs, filename, job_id):
+def read_results_json(fs, filename):
     with fs.open(filename, 'rb') as f1:
         with gzip.open(f1, 'rt', encoding='utf-8') as f2:
             dpouts = json.load(f2)
     df = pd.DataFrame(dpouts)
+    df['job_id'] = int(re.search(r'results_job(\d+)\.json\.gz', filename).group(1))
     # Sorting is needed to ensure all dfs have same column order. Dask will fail otherwise.
-    df['job_id'] = job_id
     df = df.reindex(sorted(df.columns), axis=1)
     return df
 
@@ -260,10 +260,8 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
 
     # Results "CSV"
     results_json_files = fs.glob(f'{sim_output_dir}/results_job*.json.gz')
-    job_ids = [int(re.search(r'results_job(\d+)\.json\.gz', x).group(1)) for x in results_json_files]
     if results_json_files:
-        delayed_results_dfs = [dask.delayed(read_results_json)(fs, x, job_id)
-                               for x, job_id in zip(results_json_files, job_ids)]
+        delayed_results_dfs = [dask.delayed(read_results_json)(fs, x) for x in results_json_files]
         results_df = dd.from_delayed(delayed_results_dfs)
     else:
         raise ValueError("No simulation results found to post-process.")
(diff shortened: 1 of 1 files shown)


buildstockbatch/postprocessing.py
```diff
@@ -205,12 +204,13 @@ def get_cols(fs, filename):
...
 
 
-def read_results_json(fs, filename):
+def read_results_json(fs, filename, job_id):
     with fs.open(filename, 'rb') as f1:
         with gzip.open(f1, 'rt', encoding='utf-8') as f2:
             dpouts = json.load(f2)
     df = pd.DataFrame(dpouts)
     # Sorting is needed to ensure all dfs have same column order. Dask will fail otherwise.
+    df['job_id'] = job_id
```
nmerket: This works, but wouldn't it be simpler and less error prone to parse out the job id from the filename _inside_ this function rather than to pass it in?

buildstockbatch/postprocessing.py
```diff
@@ -259,20 +259,11 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
...
-    for job_json in jobs_json_files:
-        with open(job_json, 'r') as f:
-            job_json_dict = json.load(f)
-        for building_id, upgrade in job_json_dict['batch']:
-            upgrade = 0 if upgrade is None else upgrade+1
-            job_map[upgrade][building_id] = job_json_dict['job_num']
-    logger.info(f"{len(jobs_json_files)} jobs json files gave {len(job_map)} job_map keys.")
-
     results_json_files = fs.glob(f'{sim_output_dir}/results_job*.json.gz')
+    job_ids = [int(re.search(r'results_job(\d+)\.json\.gz', x).group(1)) for x in results_json_files]
```
nmerket: see comment in the `read_results_json` function above.

nmerket (commented):
Looks pretty good. One comment below.

buildstockbatch/postprocessing.py
rajeee: Yep, it totally would be. I guess your [previous implementation](https://github.com/NREL/buildstockbatch/blob/b931a934fe540946d3bd739a0cad56209c05196f/buildstockbatch/postprocessing.py#L261) polluted my mind into thinking I should calculate them all at once! I will fix this.

nmerket: Haha. I guess so. Do as I say, not as I do. 😉

shorowit:
Ignorant question: Is it possible to create "local docker" CI tests that would catch this issue in the future? It's great that it got caught by the ResStock CI, but ideally it would be caught sooner (i.e., here).

rajeee:
> Ignorant question: Is it possible to create "local docker" CI tests that would catch this issue in the future? It's great that it got caught by the ResStock CI, but ideally it would be caught sooner (i.e., here).

@shorowit I talked about this with @joseph-robertson  as well and it seems like a good idea to have those integration tests running for buildstockbatch PR in addition to ResStock PRs. I think this can be a part of a larger discussion on overhauling the CI for BSB that @nmerket is considering.

nmerket:
@shorowit Yes, we should catch this. The way we're doing CI with Circle CI makes it hard to do, but switching to GitHub actions will make this possible. #223 should be the next thing I get to on bsb. Will chat with @joseph-robertson about best practices he's learned there.

joseph-robertson:
@nmerket @rajeee Is this ready to go?

nmerket (approved):

PR #267, opened 2022-02-08
Migrating CI to GitHub Actions

### LINKED ISSUE #223: Migrate CI to github actions

nmerket:
Fixes #223.

## Pull Request Description

description here

## Checklist

Not all may apply

- [x] ~~Tests exercising your feature/bug fix (check coverage report on CircleCI build -> Artifacts)~~
- [x] ~~Run a small batch run to make sure it all works (local is fine, unless an Eagle specific feature)~~
- [x] Update PR template
- [x] Update badge


.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 buildstockbatch --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 buildstockbatch --count --max-line-length=127 --statistics
+      - name: Run PyTest and Coverage
+        run: |
+          pytest --junitxml=coverage/junit.xml --cov=buildstockbatch --cov-report=xml:coverage/coverage.xml --cov-report=html:coverage/htmlreport
+      - name: Test Report
+        uses: mikepenz/action-junit-report@v2.9.1
+        if: ${{ matrix.python-version == '3.9' }}
```
joseph-robertson, line 31: Why Test Report thru Save Docs only for Python 3.9? Just because it's for creating coverage report?

setup.py
```diff
@@ -49,7 +49,7 @@ def run_tests(self):
     long_description_content_type='text/markdown',
     url=metadata['__url__'],
     packages=setuptools.find_packages(),
-    python_requires='>=3.6',
+    python_requires='>=3.7',
```
joseph-robertson, line 52: Plan would be to remove older versions of these from the `python-version` matrix if this gets bumped over time?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+          path: coverage/htmlreport/
+      - name: Report coverage to PR
+        uses: 5monkeys/cobertura-action@v12
+        if: ${{ matrix.python-version == '3.9' }}
+        with:
+          path: coverage/coverage.xml
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          minimum_coverage: 24
+          fail_below_threshold: true
+      - name: Build documentation
```
joseph-robertson, line 50: I don't understand why this needs to happen here? I thought docs were built automatically with https://readthedocs.org/projects/buildstockbatch/.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install buildstockbatch
+        run: |
+          python -m pip install --progress-bar off --upgrade pip
+          pip install .[dev] --progress-bar off
+      - name: Linting
```
joseph-robertson, line 20: Not required, but you could have corrections committed automatically.

joseph-robertson (approved):
Some minor comments/questions.

Maybe one general thought is: do you want to experiment with running `buildstock_docker` in this PR or in subsequent ones? I wonder if the config's jobs/structure would end up getting modified.

.github/workflows/ci.yml
nmerket: Yeah, just so the coverage report only gets created once, not 4 times.

setup.py
nmerket: Yep. We'll keep this up to date and synced with the supported versions of python. I had to drop 3.6 because the numpy dependency dropped support for python 3.6.

.github/workflows/ci.yml
nmerket: This is so people can download the generated docs from the artifact to ensure they're working right before it gets pushed to RTD. We don't use these for publishing.

.github/workflows/ci.yml
nmerket: That's interesting. I haven't done that. I actually want to switch our linting/formatting to [black](https://github.com/psf/black) but was going to do that as its own PR once a few more things have merged down because it's going to touch _a lot_ of lines of code.

nmerket:
@joseph-robertson as for your larger question about doing some runs with `buildstock_docker` I was thinking about making that part of another PR, just so we can get this one out there. I'd like to talk through with you what we should be testing here vs in the resstock repo.

PR #271, opened 2022-03-08
Apply correct schema for all upgrades

rajeee:
## Pull Request Description

When different upgrades have different set of columns with some some columns having all empty values, the datatype for the columns will not be properly set. This PR fixes the issue by scanning all the results_jobX.json files in the beginning and determines the datatype for all columns. 

## Checklist

Not all may apply


buildstockbatch/postprocessing.py
```diff
@@ -276,25 +319,38 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
...
+    if not results_json_files:
         raise ValueError("No simulation results found to post-process.")
 
+    logger.info("Collecting all the columns and datatypes in results_job*.json.gz parquet files.")
+    all_schema_dict = db.from_sequence(results_json_files).map(partial(get_schema_dict, fs)).\
+        fold(lambda x, y: merge_schema_dicts(x, y)).compute()
+    logger.info(f"Got {len(all_schema_dict)} columns")
+    all_results_cols = list(all_schema_dict.keys())
+    all_schema_dict = {to_camelcase(key): value for key, value in all_schema_dict.items()}
+    logger.info(f"Got this schema: {all_schema_dict}\n")
```
nmerket, line 331: Seems unnecessary to print this out. Maybe helpful for debugging, but now that it's working you're probably good to remove this.

nmerket, line 331: Seems unnecessary to print this out. Maybe helpful for debugging, but now that it's working you're probably good to remove this.

```suggestion
```

nmerket (commented):
Looks good. One minor suggestion.

buildstockbatch/postprocessing.py
rajeee: I debated keeping it or removing it. But I kept it because over information (upto a limit!) is better than under information for debugging, and postprocessing.out is kinda used mostly for debugging, so may it's fine to leave? Most users probably don't need to read the postprocessing.out unless they can't find their table in Athena.

nmerket (approved):

PR #273, opened 2022-03-16
New feature: Partitioning

rajeee:
## Pull Request Description

Adds partitioning capability to buildstockbatch postprocessing. And while doing that:

1. Uses dasks 'to_parquet' function to write out the parquet files along with the metadata files. This allows for almost very fast "reading"* of the timeseries files back into dask. (* It doesn't really read the whole file but it saves dask from scanning the file to infer number of columns and column datatypes). 
2. Adds n_procs argument to eagle postprocessing which defaults to 5. We have been using n_procs=1 because of some OOM scare we had, but on testing with new workflow, dask seems to be able to handle n_procs=5 quite easily, so this is the new default. Benchmark below for the difference it can make.
3. Adds more logging in postprocessing workflow. It's a bit quite a bit more verbose but I think it can be valuable for debugging.

image
Comparing the postprocessing time between n_procs=1, n_procs=5 and n_procs=10. Also comparing the upload time to s5cmd. There is a slight incentive to try to switch to s5cmd for upload, and if we do s5cmd on multiple nodes, there is a chance that the upload time can come down further. But, it's not that bad as is either. 
n_procs=10 seems to cause problem with the default 85GB memory nodes. It might work with higher memory nodes. The default for n_procs is chosen to be 5, which works for 85GB nodes.

## Checklist

Not all may apply


rajeee:
Will create a new PR.

PR #275, opened 2022-03-18
Enable partitioning

rajeee:
## Pull Request Description
Adds partitioning capability to buildstockbatch postprocessing. And while doing that:

1.U̶s̶e̶s̶ ̶d̶a̶s̶k̶s̶ ̶'̶t̶o̶_̶p̶a̶r̶q̶u̶e̶t̶'̶ ̶f̶u̶n̶c̶t̶i̶o̶n̶ ̶t̶o̶ ̶w̶r̶i̶t̶e̶ ̶o̶u̶t̶ ̶t̶h̶e̶ ̶p̶a̶r̶q̶u̶e̶t̶ ̶f̶i̶l̶e̶s̶ ̶a̶l̶o̶n̶g̶ ̶w̶i̶t̶h̶ ̶t̶h̶e̶ ̶m̶e̶t̶a̶d̶a̶t̶a̶ ̶f̶i̶l̶e̶s̶.̶ ̶T̶h̶i̶s̶ ̶a̶l̶l̶o̶w̶s̶ ̶f̶o̶r̶ ̶a̶l̶m̶o̶s̶t̶ ̶v̶e̶r̶y̶ ̶f̶a̶s̶t̶ ̶"̶r̶e̶a̶d̶i̶n̶g̶"̶*̶ ̶o̶f̶ ̶t̶h̶e̶ ̶t̶i̶m̶e̶s̶e̶r̶i̶e̶s̶ ̶f̶i̶l̶e̶s̶ ̶b̶a̶c̶k̶ ̶i̶n̶t̶o̶ ̶d̶a̶s̶k̶.̶ ̶(̶*̶ ̶I̶t̶ ̶d̶o̶e̶s̶n̶'̶t̶ ̶r̶e̶a̶l̶l̶y̶ ̶r̶e̶a̶d̶ ̶t̶h̶e̶ ̶w̶h̶o̶l̶e̶ ̶f̶i̶l̶e̶ ̶b̶u̶t̶ ̶i̶t̶ ̶s̶a̶v̶e̶s̶ ̶d̶a̶s̶k̶ ̶f̶r̶o̶m̶ ̶s̶c̶a̶n̶n̶i̶n̶g̶ ̶t̶h̶e̶ ̶f̶i̶l̶e̶ ̶t̶o̶ ̶i̶n̶f̶e̶r̶ ̶n̶u̶m̶b̶e̶r̶ ̶o̶f̶ ̶c̶o̶l̶u̶m̶n̶s̶ ̶a̶n̶d̶ ̶c̶o̶l̶u̶m̶n̶ ̶d̶a̶t̶a̶t̶y̶p̶e̶s̶)̶.̶ Manually writes dask _metadata and _common_metadata files to enable fast reading back of the aggregated dataset.
2.	Adds n_procs argument to eagle postprocessing which defaults to 16. Makes order of magnitude improvement in postprocessing speed. 
3.	Adds more logging in postprocessing workflow. It's a quite a bit more verbose but I think it can be valuable for debugging.

image

Comparing the postprocessing time between n_procs=1, 5, 10, 20, 36. Also comparing the upload time to s5cmd. There is a slight incentive to try to switch to s5cmd for upload since the upload is really the bottleneck now. And if we do s5cmd on multiple nodes, there is a chance that the upload time can come down further. But, it's not that bad as is either and I am not sure it is worth the trouble. We had used n_procs=1 before because of some OOM scare, but on testing with the latest design, the memory consumption per worker doesn't seem to ever exceed 3GB and I verified up to n_procs=36. The new default of 16 is conservatively chosen to provide ample memory to each worker. 
Check the dask memory distribution below:
image

I started the task by using a list of delayed objects and creating dask dataframe from that using from_delayed function for doing the aggregation. This worked okay when I was doing hourly run, but it started to choke when I tested on n550K_15min runs. It turns out that creating daskdatafram using from_delayed has its limitation; if we try to do any manipulation of the dataframe at all (which we need to do for aggregation /  joining / writing with partitioning etc), it ends up calling those delayed objects (behind the scene) before it can create the task graph to understand the kind of data the delayed objects would return. I don't fully understand what's going on there, but, it aligns well with what I saw: it takes forever before I see any activity on the dask dashboard, and the memory consumption of the main process where the scheduler sits keeps increasing (eventually crashing the whole thing). This SO question and answer can provide more details and context: https://stackoverflow.com/a/71531717/2105924.  

Anyway, the conclusion from all that is that creating massive dask dataframe out of essentially unformatted raw data is going to lead to problems; we should either preprocess the data to be directly readable with read_parquet or do away with massive dask df. I decided to go with the second option (since if we are going to preprocess, we might as well write the concatenated parquet directly).  One of the reasons we wanted to use dask dataframe was that it allowed us to output _metadata files when writing to parquet. The _metadata files help with super fast reading. Good news is that, I was able to write those _metadata files manually after the aggregation was completed. So, no loss there!

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -288,6 +288,16 @@ class BuildStockBatchBase(object):
         return True
 
     @staticmethod
+    def validate_postprocessing_spec(project_file):
+        cfg = get_project_configuration(project_file)  # noqa F841
+        param_option_dict, _ = BuildStockBatchBase.get_param_option_dict(project_file)
+        partition_cols = cfg.get('postprocessing',{}).get("partition_columns",[])
+        invalid_cols = [c for c in partition_cols if c not in param_option_dict.keys()]
+        if invalid_cols:
+            raise ValidationError(f"The following partition columns are not valid: {invalid_cols}")
+        return True
+
+    @staticmethod
     def validate_xor_nor_schema_keys(project_file):
         cfg = get_project_configuration(project_file)
         major, minor = cfg.get('version', __schema_version__).split('.')
@@ -301,10 +311,7 @@ class BuildStockBatchBase(object):
         return True
 
     @staticmethod
-    def validate_options_lookup(project_file):
-        """
-        Validates that the parameter|options specified in the project yaml file is available in the options_lookup.tsv
-        """
+    def get_param_option_dict(project_file):
         cfg = get_project_configuration(project_file)
         param_option_dict = defaultdict(set)
         buildstock_dir = BuildStockBatchBase.get_buildstock_dir(project_file, cfg)
@@ -326,7 +333,15 @@ class BuildStockBatchBase(object):
         except FileNotFoundError as err:
             logger.error(f"Options lookup file not found at: '{options_lookup_path}'")
             raise err
+        return param_option_dict, invalid_options_lookup_str
 
+    @staticmethod
+    def validate_options_lookup(project_file):
+        """
+        Validates that the parameter|options specified in the project yaml file is available in the options_lookup.tsv
+        """
+        cfg = get_project_configuration(project_file)
+        param_option_dict, invalid_options_lookup_str = BuildStockBatchBase.get_param_option_dict(project_file)
         invalid_option_spec_counter = Counter()
         invalid_param_counter = Counter()
         invalid_option_counter_dict = defaultdict(Counter)
@@ -420,10 +435,10 @@ class BuildStockBatchBase(object):
                     source_str_package = source_str_upgrade + ", in package_apply_logic"
                     source_option_str_list += get_all_option_str(source_str_package, upgrade['package_apply_logic'])
 
-        # FIXME: Get this working in new downselect sampler validation.
-        # if 'downselect' in cfg:
-        #     source_str = "In downselect"
-        #     source_option_str_list += get_all_option_str(source_str, cfg['downselect']['logic'])
+        if 'downselect' in cfg or "downselect" in cfg.get('sampler',{}).get('type'):
+            source_str = "In downselect"
+            logic = cfg['downselect']['logic'] if 'downselect' in cfg else cfg['sampler']['args']['logic']
+            source_option_str_list += get_all_option_str(source_str, logic)
 
         # Gather all the errors in the option_str, if any
         error_message = ''
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -291,7 +291,7 @@ class BuildStockBatchBase(object):
     def validate_postprocessing_spec(project_file):
         cfg = get_project_configuration(project_file)  # noqa F841
         param_option_dict, _ = BuildStockBatchBase.get_param_option_dict(project_file)
-        partition_cols = cfg.get('postprocessing',{}).get("partition_columns",[])
+        partition_cols = cfg.get('postprocessing', {}).get("partition_columns", [])
         invalid_cols = [c for c in partition_cols if c not in param_option_dict.keys()]
         if invalid_cols:
             raise ValidationError(f"The following partition columns are not valid: {invalid_cols}")
@@ -435,7 +435,7 @@ class BuildStockBatchBase(object):
                     source_str_package = source_str_upgrade + ", in package_apply_logic"
                     source_option_str_list += get_all_option_str(source_str_package, upgrade['package_apply_logic'])
 
-        if 'downselect' in cfg or "downselect" in cfg.get('sampler',{}).get('type'):
+        if 'downselect' in cfg or "downselect" in cfg.get('sampler', {}).get('type'):
             source_str = "In downselect"
             logic = cfg['downselect']['logic'] if 'downselect' in cfg else cfg['sampler']['args']['logic']
             source_option_str_list += get_all_option_str(source_str, logic)
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -180,14 +180,14 @@ class BuildStockBatchBase(object):
         # Convert the timeseries data to parquet
         # and copy it to the results directory
         timeseries_filepath = os.path.join(sim_dir, 'run', 'results_timeseries.csv')
-        skiprows = [1]
         # FIXME: Allowing both names here for compatibility. Should consolidate on one timeseries filename.
-        if not os.path.isfile(timeseries_filepath):
-            timeseries_filepath = os.path.join(sim_dir, 'run', 'enduse_timeseries.csv')
-            skiprows = []
-            units_dict = {}
-        else:
+        if os.path.isfile(timeseries_filepath):
             units_dict = pd.read_csv(timeseries_filepath, nrows=1).transpose().to_dict()[0]
+            skiprows = [1]
+        else:
+            timeseries_filepath = os.path.join(sim_dir, 'run', 'enduse_timeseries.csv')
+            units_dict = {}
+            skiprows = []
 
         schedules_filepath = ''
         if os.path.isdir(os.path.join(sim_dir, 'generated_files')):
@@ -216,8 +216,8 @@ class BuildStockBatchBase(object):
                 Will rename column names like End Use: Natural Gas: Range/Oven to
                 end_use__natural_gas__range_oven__kbtu to play nice with Athena
                 """
-                unit = units_dict.get(x)
-                unit = '' if not isinstance(unit, str) else unit
+                unit = units_dict.get(x)  # missing units (e.g. for time) gets nan
+                unit = unit if isinstance(unit, str) else ''  
                 sepecial_characters = [':', ' ', '/']
                 for char in sepecial_characters:
                     x = x.replace(char, '_')
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -255,7 +255,7 @@ class BuildStockBatchBase(object):
         assert(BuildStockBatchBase.validate_reference_scenario(project_file))
         assert(BuildStockBatchBase.validate_options_lookup(project_file))
         assert(BuildStockBatchBase.validate_measure_references(project_file))
-        assert(BuildStockBatchBase.validate_options_lookup(project_file))
+        assert(BuildStockBatchBase.validate_postprocessing_spec(project_file))
         logger.info('Base Validation Successful')
         return True
 
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -185,15 +185,16 @@ class BuildStockBatchBase(object):
         if not os.path.isfile(timeseries_filepath):
             timeseries_filepath = os.path.join(sim_dir, 'run', 'enduse_timeseries.csv')
             skiprows = []
-            units_dict = None
+            units_dict = {}
         else:
-            units_dict = pd.read_csv("results_timeseries.csv", nrows=1, header=[0]).transpose().to_dict()[0]
+            units_dict = pd.read_csv("results_timeseries.csv", nrows=1).transpose().to_dict()[0]
 
         schedules_filepath = ''
         if os.path.isdir(os.path.join(sim_dir, 'generated_files')):
             for file in os.listdir(os.path.join(sim_dir, 'generated_files')):
                 if file.endswith('schedules.csv'):
                     schedules_filepath = os.path.join(sim_dir, 'generated_files', file)
+
         if os.path.isfile(timeseries_filepath):
             # Find the time columns present in the enduse_timeseries file
             possible_time_cols = ['time', 'Time', 'TimeDST', 'TimeUTC']
@@ -202,13 +203,28 @@ class BuildStockBatchBase(object):
             if not actual_time_cols:
                 logger.error(f'Did not find any time column ({possible_time_cols}) in {timeseries_filepath}.')
                 raise RuntimeError(f'Did not find any time column ({possible_time_cols}) in {timeseries_filepath}.')
-            
+
             tsdf = pd.read_csv(timeseries_filepath, parse_dates=actual_time_cols, skiprows=skiprows)
             if os.path.isfile(schedules_filepath):
                 schedules = pd.read_csv(schedules_filepath)
                 schedules.rename(columns=lambda x: f'schedules_{x}', inplace=True)
                 schedules['TimeDST'] = tsdf['Time']
                 tsdf = tsdf.merge(schedules, how='left', on='TimeDST')
+
+            def get_clean_column_name(x):
+                """"
+                Will rename column names like End Use: Natural Gas: Range/Oven to
+                end_use__natural_gas__range_oven__kbtu to play nice with Athena
+                """
+                unit = units_dict.get(x)
+                unit = '' if not isinstance(unit, str) else unit
+                sepecial_characters = [':', ' ', '/']
+                for char in sepecial_characters:
+                    x = x.replace(char, '_')
+                x = x + "__" + unit if unit else x
+                return x.lower()
+
+            tsdf.rename(columns=get_clean_column_name, inplace=True)
             postprocessing.write_dataframe_as_parquet(
                 tsdf,
                 dest_fs,
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -217,7 +217,7 @@ class BuildStockBatchBase(object):
                 end_use__natural_gas__range_oven__kbtu to play nice with Athena
                 """
                 unit = units_dict.get(x)  # missing units (e.g. for time) gets nan
-                unit = unit if isinstance(unit, str) else ''  
+                unit = unit if isinstance(unit, str) else ''
                 sepecial_characters = [':', ' ', '/']
                 for char in sepecial_characters:
                     x = x.replace(char, '_')
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -451,6 +451,7 @@ class BuildStockBatchBase(object):
                     source_str_package = source_str_upgrade + ", in package_apply_logic"
                     source_option_str_list += get_all_option_str(source_str_package, upgrade['package_apply_logic'])
 
+        #  TODO: refactor this into Sampler.validate_args
         if 'downselect' in cfg or "downselect" in cfg.get('sampler', {}).get('type'):
             source_str = "In downselect"
             logic = cfg['downselect']['logic'] if 'downselect' in cfg else cfg['sampler']['args']['logic']
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -179,14 +179,16 @@ class BuildStockBatchBase(object):
 
         # Convert the timeseries data to parquet
         # and copy it to the results directory
-        results_timeseries_filepath = os.path.join(sim_dir, 'run', 'results_timeseries.csv')
-        timeseries_filepath = results_timeseries_filepath
+        timeseries_filepath = os.path.join(sim_dir, 'run', 'results_timeseries.csv')
         skiprows = [1]
         # FIXME: Allowing both names here for compatibility. Should consolidate on one timeseries filename.
-        if not os.path.isfile(results_timeseries_filepath):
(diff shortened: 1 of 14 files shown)


buildstockbatch/schemas/v0.3.yaml
```diff
@@ -47,8 +47,10 @@ hpc-postprocessing-spec:
   time: int(required=True)
   n_workers: int(required=False)
   node_memory_mb: enum(85248, 180224, 751616, required=False)
+  n_procs: enum(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, required=False)
```
nmerket:
```suggestion
  n_procs: int(min=1, max=10, required=False)
```

see https://github.com/23andMe/Yamale#integer---intminint-maxint

buildstockbatch/postprocessing.py
```diff
@@ -292,6 +303,81 @@ def correct_schema(cur_schema_dict, df):
...
+
+def get_upgrade_list(cfg):
+    upgrade_start = 1 if cfg['baseline'].get('skip_sims', False) else 0
+    upgrade_end = len(cfg.get('upgrades', [])) + 1
+    full_upgrade_list = range(upgrade_start, upgrade_end)
+    upgrade_list = cfg.get('postprocessing', {}).get('upgrades', [])
+    upgrade_list = sorted(upgrade_list) or full_upgrade_list
+    for upgrade in upgrade_list:
+        if upgrade not in full_upgrade_list:
+            logger.warning(f"Upgrade {upgrade} specified in the yaml is not valid")
```
nmerket: I'm a little confused here. Wouldn't we want it to error out if an upgrade is missing rather than just rolling on by?

buildstockbatch/postprocessing.py
```diff
@@ -348,14 +434,25 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
...
+    upgrade_list = get_upgrade_list(cfg)
+    partition_columns = cfg.get('postprocessing', {}).get('partition_columns', [])
+    partition_columns = [c.lower() for c in partition_columns]
+    df_partition_columns = [f'build_existing_model.{c}' for c in partition_columns]
+    missing_cols = set(df_partition_columns) - set(all_schema_dict.keys())
+
+    if missing_cols:
+        df_partition_columns = list(set(df_partition_columns) - set(missing_cols))
+        logger.info(f"The following patition columns are not available on the results.json: {missing_cols}")
+        logger.info(f"Only using these as the partition column:s {df_partition_columns}")
```
nmerket: Also seems like this should error out. This should probably be a pre run validation if possible. And if we do keep this, aren't the order of the partition columns important? (i.e. state then county) This is using `set` which can return the columns in any order.

buildstockbatch/postprocessing.py
```diff
@@ -437,17 +561,13 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
...
-            )
-            group_ids = list(range(npartitions))
             with tempfile.TemporaryDirectory() as tmpdir:
                 tmpfilepath = Path(tmpdir, 'dask-report.html')
                 with performance_report(filename=str(tmpfilepath)):
-                    dask.compute(map(read_and_concat_ts_pq_d, ts_files_in_each_partition, group_ids))
+                    full_df.to_parquet(ts_out_loc,
+                                       partition_on=partition_columns,
+                                       write_index=True,
+                                       name_function=lambda i: f"group{i}.parquet")
```
nmerket: Cool. I didn't know you could give a name function like that. Also, nice job on writing from a dask dataframe.

nmerket (commented):
I think it's looking really good. I'm not sure I 100% followed everything. A few comments and suggestions below.

buildstockbatch/schemas/v0.3.yaml
rajeee: Accepted.

buildstockbatch/postprocessing.py
rajeee: I removed ability to provide custom list of upgrades (which wasn't there to begin with!). If some upgrades fail (say timeout in the middle, which is unlikely now), you would have to start over. This isn't too bad since with 32 nodes, even 550K run with several upgrades totalling 2TB of data takes only 30 minutes to aggregate. The juice is no longer worth the squeeze anymore!

buildstockbatch/postprocessing.py
rajeee: I have included validation for this, so, if the input partition_columns specified in the yaml is wrong, it will fail validation. The check here is for cases when the columns are missing somehow in the results_json (for example all sims failed or some bugs in the simulation / resstock). In that case, I think you are right, we just raise an error and crash instead of silently continuing on.

buildstockbatch/postprocessing.py
rajeee: Well, it was nice when it lasted. This no longer works when dealing with massive dataframe!

buildstockbatch/base.py
```diff
@@ -179,19 +179,22 @@ def cleanup_sim_dir(sim_dir, dest_fs, simout_ts_dir, upgrade_id, building_id):
...
-            enduse_timeseries_filepath = os.path.join(sim_dir, 'run', 'enduse_timeseries.csv')
-            timeseries_filepath = enduse_timeseries_filepath
-            skiprows = False
+        if os.path.isfile(timeseries_filepath):
+            units_dict = pd.read_csv(timeseries_filepath, nrows=1).transpose().to_dict()[0]
+            skiprows = [1]
+        else:
+            timeseries_filepath = os.path.join(sim_dir, 'run', 'enduse_timeseries.csv')
+            units_dict = {}
+            skiprows = []
```
nmerket, line 190: Does this mean that for older runs with the "enduse_timeseries.csv" files, it won't be able to do the metadata stuff? If so, maybe we should just remove support for the older filenames.

buildstockbatch/base.py
```diff
@@ -239,7 +255,7 @@ def validate_project(project_file):
         assert(BuildStockBatchBase.validate_reference_scenario(project_file))
         assert(BuildStockBatchBase.validate_options_lookup(project_file))
         assert(BuildStockBatchBase.validate_measure_references(project_file))
-        assert(BuildStockBatchBase.validate_options_lookup(project_file))
+        assert(BuildStockBatchBase.validate_postprocessing_spec(project_file))
```
nmerket, line 258: Do we not want to validate options lookup anymore? Is that now covered in the validate postprocessing spec?

buildstockbatch/base.py
```diff
@@ -420,10 +451,10 @@ def get_all_option_str(source_str, inp):
...
                     source_option_str_list += get_all_option_str(source_str_package, upgrade['package_apply_logic'])
 
-        # FIXME: Get this working in new downselect sampler validation.
-        # if 'downselect' in cfg:
-        #     source_str = "In downselect"
-        #     source_option_str_list += get_all_option_str(source_str, cfg['downselect']['logic'])
+        if 'downselect' in cfg or "downselect" in cfg.get('sampler', {}).get('type'):
+            source_str = "In downselect"
+            logic = cfg['downselect']['logic'] if 'downselect' in cfg else cfg['sampler']['args']['logic']
+            source_option_str_list += get_all_option_str(source_str, logic)
```
nmerket: Thanks for addressing this. I think I meant to push some of this down to the workflow generator validation. This is a little hacky just searching for "downselect" in the config. I could see this breaking easily.

buildstockbatch/postprocessing.py
```diff
@@ -214,8 +216,9 @@ def clean_up_results_df(df, cfg, keep_upgrade_id=False):
     return results_df
 
 
-def get_cols(fs, filename):
-    with fs.open(filename, 'rb') as f:
+def get_cols(fs, path, filename):
+    filepath = Path(path) / filename
+    with fs.open(filepath, 'rb') as f:
```
nmerket: Yeah, this could break things depending on which platform you're using. (I'm thinking Windows with the `\` instead of the `/`.) For compatibility with using with on S3, we need to have all file access be done through the [fsspec](https://filesystem-spec.readthedocs.io/en/latest/index.html) object `fs`. Path joining is a [known deficiency](https://github.com/fsspec/filesystem_spec/issues/747) with that library. It'd probably be best to just do this as a string. 

```suggestion
    filepath = f"{path}/{filename}"
    with fs.open(filepath, 'rb') as f:
```

buildstockbatch/postprocessing.py
```diff
@@ -251,22 +254,41 @@ def merge_schema_dicts(dict1, dict2):
...
-    return df[all_cols]
+    df = df[all_cols]
+    df.set_index('building_id', inplace=True)
+    return df
+
+
+def concat_and_normalize(fs, all_cols, src_path, dst_path, partition_columns, indx, bldg_ids, partition_vals):
+    dfs = []
+    for bldg_id in sorted(bldg_ids):
+        src_filename = Path(src_path) / f"bldg{bldg_id:07}.parquet"
```
nmerket:
```suggestion
        src_filename = f"{src_path)}/bldg{bldg_id:07}.parquet"
```

buildstockbatch/postprocessing.py
```diff
@@ -251,22 +254,41 @@ def merge_schema_dicts(dict1, dict2):
...
+        df = pd.read_parquet(src_filename, engine='pyarrow')
```
nmerket: Should use `fs`, something like this.

```suggestion
        with fs.open(src_filename, 'rb') as f:
            df = pd.read_parquet(f, engine='pyarrow')
```

buildstockbatch/postprocessing.py
```diff
@@ -251,22 +254,41 @@ def merge_schema_dicts(dict1, dict2):
...
+        df = df[all_cols]
+        df.set_index('building_id', inplace=True)
+        dfs.append(df)
+    df = pd.concat(dfs)
+    del dfs
 
+    dst_filepath = Path(dst_path)
+    for c, v in zip(partition_columns, partition_vals):
+        folder_name = f"{c}={v}"
+        dst_filepath = dst_filepath / folder_name
```
nmerket:
```suggestion
    for c, v in zip(partition_columns, partition_vals):
        dst_filepath = f"{dst_path}/{c}={v}"
```

buildstockbatch/postprocessing.py
```diff
@@ -251,22 +254,41 @@ def merge_schema_dicts(dict1, dict2):
...
 
-def read_and_concat_enduse_timeseries_parquet(fs, all_cols, output_dir, filenames, group_id):
-    dfs = [read_enduse_timeseries_parquet(fs, filename, all_cols) for filename in filenames]
-    grouped_df = pd.concat(dfs)
-    grouped_df.set_index('building_id', inplace=True)
-    grouped_df.to_parquet(output_dir + f'group{group_id}.parquet')
-    del grouped_df
+    fs.makedirs(dst_filepath, exist_ok=True)
+    dst_filename = dst_filepath / f"group{indx}.parquet"
+    df.to_parquet(dst_filename, index=True)
```
nmerket:
```suggestion
    fs.makedirs(dst_filepath, exist_ok=True)
    dst_filename = f"{dst_filepath}/group{indx}.parquet"
    with fs.open(dst_filename, 'wb') as f:
        df.to_parquet(f, index=True)
```

buildstockbatch/postprocessing.py
```diff
@@ -292,6 +314,88 @@ def correct_schema(cur_schema_dict, df):
     return sch, unresolved
 
 
+def split_into_groups(total_size, max_group_size):
+    if total_size == 0:
+        return []
+    total_groups = math.ceil(total_size / max_group_size)
+    min_elements_per_group = math.floor(total_size / total_groups)
+    split_array = [min_elements_per_group]*total_groups
```
nmerket:
```suggestion
    split_array = [min_elements_per_group] * total_groups
```

buildstockbatch/postprocessing.py
```diff
@@ -292,6 +314,88 @@ def correct_schema(cur_schema_dict, df):
...
+        cur_index = indx
+
+    return bldg_id_groups, bldg_id_list, ngroups
+
+
+def get_bldg_groups(bldg_id_list, files_per_partition):
+    """
+        Similar to get_partitioned_buildings_groups, but without the partition columns. Groups buildings into sub list
+        of max of files_per_partition per group.
+    """
```
nmerket: This seems kind of redundant. Is there any way this could be combined with `get_partitioned_bldg_groups` and just have it work with `partition_columns = []`?

buildstockbatch/postprocessing.py
```diff
@@ -292,6 +314,88 @@ def correct_schema(cur_schema_dict, df):
...
+    return list(range(upgrade_start, upgrade_end))
+
+
+def write_metadata_files(fs, parquet_root_dir, partition_columns):
+    df = dd.read_parquet(parquet_root_dir)
+    sch = pa.Schema.from_pandas(df._meta_nonempty)
+    parquet.write_metadata(sch, Path(parquet_root_dir) / "_common_metadata")
+    logger.info(f"Written _common_metadata to {parquet_root_dir}")
+    partition_glob = Path(*[f'{c}*' for c in partition_columns]) if partition_columns else ""
+    glob_str = Path(parquet_root_dir) / "up*" / partition_glob / "*.parquet"
```
nmerket: Remove `Path` references here too before using `fs.glob`.

buildstockbatch/postprocessing.py
```diff
@@ -315,7 +419,7 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
 
     # create the postprocessing results directories
     for dr in dirs:
-        fs.makedirs(dr)
+        fs.makedirs(dr, exist_ok=True)
```
nmerket: Interesting that you're running into situations where the directories already exist here. This was to pre-create all the dirs so that we didn't have race conditions on directory creation further down.

buildstockbatch/postprocessing.py
```diff
@@ -405,53 +530,66 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
...
-            total_mem = mean_mem * len(ts_filenames) / 1e6  # total_mem in MB
 
             # Determine how many files should be in each partition and group the files
             parquet_memory = int(cfg['eagle'].get('postprocessing', {}).get('parquet_memory_mb', MAX_PARQUET_MEMORY))
             logger.info(f"Max parquet memory: {parquet_memory} MB")
-            npartitions = math.ceil(total_mem / parquet_memory)
-            npartitions = min(len(ts_filenames), npartitions)  # cannot have less than one file per partition
-            ts_files_in_each_partition = np.array_split(ts_filenames, npartitions)
+            max_files_per_partition = max(1, math.floor(parquet_memory / (mean_mem / 1e6)))
+            ts_bldg_ids = [int(f[-15:-8]) for f in ts_filenames]
```
nmerket: Could this be done with a regex so it's a little clearer what number you're pulling out of the path?

buildstockbatch/postprocessing.py
```diff
@@ -471,14 +609,16 @@ def upload_results(aws_conf, output_dir, results_dir):
 
     output_folder_name = Path(output_dir).name
     parquet_dir = Path(results_dir).joinpath('parquet')
-
+    ts_dir = parquet_dir / 'timeseries'
```
nmerket, line 598: And I'm reminding myself that we're okay to use pathlib here because this only runs locall to put files on s3 that aren't already there, so it isn't used in the AWS implementation.

buildstockbatch/postprocessing.py
```diff
@@ -528,7 +668,7 @@ def create_athena_tables(aws_conf, tbl_prefix, s3_bucket, s3_prefix):
     crawlTarget = {
         'S3Targets': [{
             'Path': s3_path,
-            'Exclusions': []
+            'Exclusions': ['**_metadata', '**_common_metadata']
         }]
     }
     crawler_name = db_name + '_' + tbl_prefix
```
nmerket, line 660: This shouldn't be part of this PR, but you can create a glue table directly now based on a pandas dataframe schema using AWS data wrangler. We could skip needing a crawler altogether. I'll make an issue for this.

buildstockbatch/schemas/v0.3.yaml
```diff
@@ -47,8 +47,10 @@ hpc-postprocessing-spec:
   time: int(required=True)
   n_workers: int(required=False)
   node_memory_mb: enum(85248, 180224, 751616, required=False)
+  n_procs: int(min=1, max=32, required=False)
```
nmerket: Let's make sure this is documented well. It could be confusing for people differentiating `n_procs` and `n_workers`.

buildstockbatch/test/test_base.py
```diff
@@ -320,6 +323,16 @@ def test_upload_files(mocked_boto3, basic_residential_project_file):
...
+    s3_file_path = s3_path + 'timeseries/_common_metadata'
+    source_file_path = os.path.join(source_path, 'timeseries', '_common_metadata')
+    assert (source_file_path, s3_file_path) in files_uploaded
+    files_uploaded.remove((source_file_path, s3_file_path))
+
+    s3_file_path = s3_path + 'timeseries/_metadata'
+    source_file_path = os.path.join(source_path, 'timeseries', '_metadata')
+    assert (source_file_path, s3_file_path) in files_uploaded
+    files_uploaded.remove((source_file_path, s3_file_path))
+
```
nmerket, line 335: As I was reviewing this, I had an idea for future testing. There are services that mock s3 locally. We should have a set of tests that runs the postprocessing on s3fs to ensure we don't break in the future.

docs/changelog/changelog_dev.rst
```diff
@@ -38,8 +38,17 @@ Development Changelog
         non-missing values.
 
     .. change::
+        :tags: postprocessing, feature
+        :pullreq: 275
+        :tickets:
+
+        Postprocessing can take partition the data before uploading to s3 and Athena. This allows for faster and cheaper
+        queries.
```
nmerket: Mention the input yaml changes you made here too.

nmerket (commented):
See my comments and notes below. A lot of what needs to be fixed is removing pathlib from when we're interacting with the fsspec filesystem. I tried to make suggestions on how to do that but probably missed a few places. I also don't guarantee the suggestions work. 

Overall, really nice work!

buildstockbatch/base.py
rajeee: There was a duplicate line. See few lines above.

buildstockbatch/postprocessing.py
rajeee: Ha, I thought I was being more careful by using pathlibs path joining than using string "/" join.

buildstockbatch/base.py
rajeee: Yeah, agreed it's a bit hacky. Since we are already doing apply_logic validation here just thought I would uncomment and repair existing test. But, yeah, I would support refactoring the logic validation piece and then calling downselect logic validation from within the sampler.

buildstockbatch/base.py
rajeee: Not really. The older enduse_timeseries.csv have units already baked into the column names ([kWh] suffix) so no need to be clever to find the units. The new version has column units in the second row. This is unrelated to writing _metadata files.

buildstockbatch/postprocessing.py
rajeee: I will revert this. I added this because at one point I was working on allowing to resume from last failed upgrade. I have removed that capability now (if you need to redo postprocessing, you redo all).

buildstockbatch/postprocessing.py
rajeee: Good point. I have refactored this.

buildstockbatch/base.py
rajeee: Add Todo: to refactor this into the sampler validation.

nmerket (approved):
Thanks for addressing all my nit picking.

PR #274, opened 2022-03-18
HPXML workflow generator documentation updates

joseph-robertson:
## Pull Request Description

- [x] Move any measures specified in `measures` to _after_ HPXMLtoOpenStudio
- [x] Improve validation from the residential hpxml workflow generator
- [x] Add a bunch of missing measure argument descriptions to the residential hpxml workflow generator docs

## Checklist

Not all may apply

- [ ] Code changes (must work)
- [ ] Tests exercising your feature/bug fix (check coverage report on Checks -> BuildStockBatch Tests -> Artifacts)
- [ ] Coverage has increased or at least not decreased. Update `minimum_coverage` in `.github/workflows/ci.yml` as necessary.
- [ ] All other unit tests passing
- [ ] Update validation for project config yaml file changes
- [ ] Update existing documentation
- [ ] Run a small batch run to make sure it all works (local is fine, unless an Eagle specific feature)
- [ ] Add to the changelog_dev.rst file and propose migration text in the pull request

PR #280, opened 2022-03-24
Allow timeseries total site energy use for residential_hpxml

shorowit:
## Pull Request Description

Adds support for new `include_timeseries_total_consumptions` argument of the ReportSimulationOutput measure.

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/workflow_generator/residential_hpxml.py
+++ b/buildstockbatch/workflow_generator/residential_hpxml.py
@@ -107,6 +107,7 @@ class ResidentialHpxmlWorkflowGenerator(WorkflowGeneratorBase):
 
         sim_out_rep_args = {
             'timeseries_frequency': 'none',
+            'include_timeseries_total_consumptions': False,
             'include_timeseries_fuel_consumptions': False,
             'include_timeseries_end_use_consumptions': True,
             'include_timeseries_emissions': False,
--- a/docs/workflow_generators/residential_hpxml.rst
+++ b/docs/workflow_generators/residential_hpxml.rst
@@ -23,6 +23,7 @@ Configuration Example
 
         simulation_output_report:
           timeseries_frequency: hourly
+          include_timeseries_total_consumptions: true
           include_timeseries_fuel_consumptions: true
           include_timeseries_end_use_consumptions: true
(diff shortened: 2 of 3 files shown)


docs/workflow_generators/residential_hpxml.rst
```diff
@@ -73,9 +74,9 @@ Arguments
   Note that the default behavior is to retain some files and remove others.
   See :ref:`server-dir-cleanup-defaults` for current defaults.
 
-.. _BuildExistingModel: https://github.com/NREL/resstock/blob/restructure-v3/measures/BuildExistingModel/measure.xml
-.. _ReportSimulationOutput: https://github.com/NREL/resstock/blob/restructure-v3/resources/hpxml-measures/ReportSimulationOutput/measure.xml
-.. _ServerDirectoryCleanup: https://github.com/NREL/resstock/blob/restructure-v3/measures/ServerDirectoryCleanup/measure.xml
+.. _BuildExistingModel: https://github.com/NREL/resstock/blob/develop/measures/BuildExistingModel/measure.xml
```
joseph-robertson, line 77: Good catch, although caught here: https://github.com/NREL/buildstockbatch/pull/274

buildstockbatch/workflow_generator/residential_hpxml.py
```diff
@@ -107,6 +107,7 @@ def create_osw(self, sim_id, building_id, upgrade_idx):
 
         sim_out_rep_args = {
             'timeseries_frequency': 'none',
+            'include_timeseries_total_consumptions': False,
```
joseph-robertson, line 110: Should we default to `True`?

joseph-robertson (approved):

buildstockbatch/workflow_generator/residential_hpxml.py
shorowit: `include_timeseries_fuel_consumptions` is being defaulted to false, so this seemed consistent with that. The default is to get all end uses and no level of aggregation, which I think makes sense as it prevents a user from accidentally summing across all columns and not realizing there are aggregated outputs.

PR #294, opened 2022-06-03
Anticipate 0.22 as next version

joseph-robertson:
Fixes # .

## Pull Request Description

Update `__version__` to 0.22.

Create a `validate_resstock_version` method for checking current BuildstockBatch version against version required by ResStock.

Create a `validate_openstudio_version` method for checking current OpenStudio version against version required by OS-HPXML.

Pass `BUILDSTOCKBATCH_VERSION` env var through for eagle and localdocker (used by ResStock for checking version, which is useful for people trying to use old BuildstockBatch with new ResStock).

Companion PR: https://github.com/NREL/resstock/pull/938.

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -588,25 +588,27 @@ class BuildStockBatchBase(object):
     @staticmethod
     def validate_resstock_version(project_file):
         """
-        Check
+        Checks the minimum required version of BuildStockBatch against the version being used
         """
         cfg = get_project_configuration(project_file)
 
-        versions = {}
-        with open(os.path.join(cfg['buildstock_directory'], 'resources/buildstock.rb'), 'r') as f:
-            for line in f:
-                line = line.strip()
-                for tool in ['ResStock_Version', 'BuildStockBatch_Version']:
-                    if line.startswith(tool):
-                        lhs, rhs = line.split('=')
-                        version, _ = rhs.split('#')
-                        versions[tool] = eval(version.strip())
-        ResStock_Version = versions['ResStock_Version']
-        BuildStockBatch_Version = versions['BuildStockBatch_Version']
-        if bsb_version < BuildStockBatch_Version:
-            val_err = f"BuildStockBatch version {BuildStockBatch_Version} or above is required" \
-                f" for ResStock version {ResStock_Version}"
-            raise ValidationError(val_err)
+        buildstock_rb = os.path.join(cfg['buildstock_directory'], 'resources/buildstock.rb')
+        if os.path.exists(buildstock_rb):
+            versions = {}
+            with open(buildstock_rb, 'r') as f:
+                for line in f:
+                    line = line.strip()
+                    for tool in ['ResStock_Version', 'BuildStockBatch_Version']:
+                        if line.startswith(tool):
+                            lhs, rhs = line.split('=')
+                            version, _ = rhs.split('#')
+                            versions[tool] = eval(version.strip())
+            ResStock_Version = versions['ResStock_Version']
+            BuildStockBatch_Version = versions['BuildStockBatch_Version']
+            if bsb_version < BuildStockBatch_Version:
+                val_err = f"BuildStockBatch version {BuildStockBatch_Version} or above is required" \
+                    f" for ResStock version {ResStock_Version}"
+                raise ValidationError(val_err)
 
         return True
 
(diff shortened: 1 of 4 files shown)


nmerket (approved):

rajeee:
@joseph-robertson  
Is it possible to also validate the OS version requirement? Like add minimum OS version required in ResStock along side BSB version requirement (resources/buildstock.rb) and on the BSB yaml validator check that the OS version being used is correct.

joseph-robertson:
The first is already being done (via OS-HPXML). For the second, doesn't that conflict with what we talked about regarding `os_version` and `os_sha` (see description [here](https://github.com/NREL/resstock/pull/938)). Maybe it doesn't.

@rajeee

rajeee:
@joseph-robertson  I don't think it does. It will be companion to that. In that PR you have specified the OS version and SHA in they "example" project yaml file. But, nothing really prevents people from attempting to run without specifying those OS version (or specifying wrong version) in your own project file. I wanted to have that addressed via the validation. 
Since the OS version is only mentioned in the yaml file (and people might change/override them), the validator has no reference to check against. If we add OS version alongside BuildStockBatch_Version, then the validator can check if they OS version used in they yaml is bigger than OS_Version in buildstock.rb or not.

buildstockbatch/base.py
```diff
@@ -583,6 +585,33 @@ def validate_reference_scenario(project_file):
...
+    def validate_resstock_version(project_file):
+        """
+        Checks the minimum required version of BuildStockBatch against the version being used
+        """
+        cfg = get_project_configuration(project_file)
+
+        buildstock_rb = os.path.join(cfg['buildstock_directory'], 'resources/buildstock.rb')
+        if os.path.exists(buildstock_rb):
+            versions = {}
+            with open(buildstock_rb, 'r') as f:
```
rajeee, line 599: I kinda feel uneasy about string searching the buildstock.rb file to find out the ResStock_Version and BuildStockBatch_Version, but don't really know a better way!

joseph-robertson:
See [commit above](https://github.com/NREL/buildstockbatch/pull/294/commits/5600c3702fa912934434e8f04bc61cdbff843292).

buildstockbatch/base.py
```diff
@@ -612,6 +613,31 @@ def validate_resstock_version(project_file):
...
+            with open(version_rb, 'r') as f:
+                for line in f:
+                    line = line.strip()
+                    for tool in ['OS_Version']:
+                        if line.startswith(tool):
+                            lhs, rhs = line.split('=')
+                            version, _ = rhs.split('#')
+                            versions[tool] = eval(version.strip())
+            OS_Version = versions['OS_Version']
+            if cfg['os_version'] != OS_Version:
```
rajeee: cfg may not have os_version (it's optional), in which case it should use the default version. (Maybe we can specify the default in the yaml schema?)
Also, don't we want inequality comparison? If cfg version is greater than OS_Version, wouldn't it be fine?

joseph-robertson: Actually, no. OS-HPXML requires an exact version, not a minimum version: https://github.com/NREL/OpenStudio-HPXML/blob/master/HPXMLtoOpenStudio/resources/version.rb#L8-L16.

rajeee: Okay, sounds good then. Might still need to handle missing "OS_Version" in the yaml though.

joseph-robertson: Yep, got that in.

buildstockbatch/base.py
```diff
@@ -620,8 +620,10 @@ def validate_openstudio_version(project_file):
         """
         cfg = get_project_configuration(project_file)
 
-        version_rb = os.path.join(cfg['buildstock_directory'],
-            'resources/hpxml-measures/HPXMLtoOpenStudio/resources/version.rb')
+        if not 'os_version' in cfg:
+            return True
```
rajeee: This can be a problem. BSB will use version from here if it is not provided in config. 
https://github.com/NREL/buildstockbatch/blob/515506f7a2fb1e5abb4677c2bcd27631b09271bf/buildstockbatch/base.py#L42

So, maybe use something like this:
os_version = cfg.get('os_version', BuildStockBatchBase.DEFAULT_OS_VERSION)

rajeee (approved):
Looks good now.
Thanks for adding in this feature.

Issue #254, opened 2021-12-28
Save YAML file on Eagle and AWS


afontani:
**Is your feature request related to a problem? Please describe.**
It would be nice to have the YAML file copied to Eagle and uploaded to S3 so that we know more about the results, the upgrade scenarios, where data might be on Eagle/AWS, and possibly track down the owner.

**Describe the solution you'd like**
In the workflow generator copy the YAML file to the outupt directory on Eagle. In the upload script also upload the YAML file to AWS.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #301, opened 2022-07-19
Add period beginning feature

rajeee:
## Pull Request Description

Currently, OS-HPXML outputs timestamps in period ending format. This PR enables converting them to period beginning format. We require period beginning for LA100ES project. 

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -171,7 +171,7 @@ class BuildStockBatchBase(object):
             diff = df['timedst'] - df['time']
             offset_diff = np.array([0] + list(diff.values)[:-1])
             df['time'] = df['time'] - step
-            df['timedst'] = df['time'] + np.array(offset_diff)
+            df['timedst'] = df['time'] + offset_diff
         else:
             df['time'] = df['time'] - step
 
@@ -244,7 +244,7 @@ class BuildStockBatchBase(object):
                 return x.lower()
 
             tsdf.rename(columns=get_clean_column_name, inplace=True)
-            if cfg.get("postprocessing", {}).get("timestamps_uses_period_beginning", False):
+            if cfg.get("postprocessing", {}).get("timestamps_use_period_beginning", False):
                 BuildStockBatchBase.make_period_beginning(tsdf)
             postprocessing.write_dataframe_as_parquet(
                 tsdf,
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -180,8 +180,8 @@ class BuildStockBatchBase(object):
 
         return df
 
-    @staticmethod
-    def cleanup_sim_dir(cfg, sim_dir, dest_fs, simout_ts_dir, upgrade_id, building_id):
+    @classmethod
+    def cleanup_sim_dir(cls, cfg, sim_dir, dest_fs, simout_ts_dir, upgrade_id, building_id):
         """Clean up the output directory for a single simulation.
 
         :param sim_dir: simulation directory
@@ -245,7 +245,7 @@ class BuildStockBatchBase(object):
 
             tsdf.rename(columns=get_clean_column_name, inplace=True)
             if cfg.get("postprocessing", {}).get("timestamps_use_period_beginning", False):
-                BuildStockBatchBase.make_period_beginning(tsdf)
+                cls.make_period_beginning(tsdf)
             postprocessing.write_dataframe_as_parquet(
                 tsdf,
                 dest_fs,
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -16,6 +16,7 @@ from fsspec.implementations.local import LocalFileSystem
 import logging
 import os
 import pandas as pd
+import numpy as np
 import requests
 import shutil
 import tempfile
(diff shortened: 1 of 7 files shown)


joseph-robertson:
I wonder if this would be more efficiently handled (optionally) on the OS-HPXML side...

shorowit:
To be clear, OS-HPXML just outputs whatever EnergyPlus outputs, which is indeed period ending timestamps. I assume outputs from ComStock are the same. It seems like it could cause confusion to stray from the EnergyPlus convention...?

buildstockbatch/base.py
```diff
@@ -164,7 +165,23 @@ def make_sim_dir(building_id, upgrade_idx, base_dir, overwrite_existing=False):
...
 
     @staticmethod
-    def cleanup_sim_dir(sim_dir, dest_fs, simout_ts_dir, upgrade_id, building_id):
+    def make_period_beginning(df):
+        step = df['time'].iloc[1] - df['time'].iloc[0]
+        if 'timedst' in df.columns:
+            diff = df['timedst'] - df['time']
+            offset_diff = np.array([0] + list(diff.values)[:-1])
+            df['time'] = df['time'] - step
+            df['timedst'] = df['time'] + np.array(offset_diff)
```
afontani: L174 is a good solution to the daylight savings shifting in timestamps.

buildstockbatch/schemas/v0.3.yaml
```diff
@@ -109,6 +109,7 @@ postprocessing-spec:
   partition_columns: list(str(), required=False)
   aws: include('aws-postprocessing-spec', required=False)
   keep_individual_timeseries: bool(required=False)
+  timestamps_uses_period_beginning: bool(required=False)
```
afontani: Just to be clear, is this the proper input in the YAML?
```
postprocessing:
  timestamps_uses_period_beginning: true
  aws:
    ...
```

afontani (commented):
@rajeee The implementation of the feature makes sense to me. I will leave it up to @nmerket if this warrants a merge with `develop`. If we want to merge this into `develop`, we should write some tests making sure that the shifting is done correctly especially around daylight savings times and the end of year timestamp.

afontani:
> To be clear, OS-HPXML just outputs whatever EnergyPlus outputs, which is indeed period ending timesteps. I assume outputs from ComStock are the same. It seems like it could cause confusion to stray from the EnergyPlus convention...?

@shorowit, it could definitely get confusing. This capability has been wanted when we connect with other tools and teams. Examples are the LA100, LA100 ES, and dsgrid projects. In these cases, we as a larger group of modeling teams have to come to a consensus about either period ending or period beginning timestamps. In these cases, we have been forced to post-process our data to conform to the needs of other teams (shifting our energy timeseries to period beginning timestamps).

buildstockbatch/schemas/v0.3.yaml
rajeee: Yes.

afontani:
I did a comparison between the timestamps before and after with `buildstock_docker`. I used 2012 as a comparison point. The daylight savings dates for 2012 are March 11 - November 4.
timestep-comparison
.

rajeee:
@nmerket @shorowit 
What do you think of merging in this feature to develop? If you think it is worthwhile to merge this to develop, I will spend some time to write tests for this. If you think this feature is too project specific, I will not bother.

nmerket:
I think period beginning timestamps are better for a variety of reasons. However, I do have concerns about unanticipated downstream implications of this change. I like having it available to anyone as long as its optional and turned off by default. But yes, I do think some tests are in order as there are a lot of weird edge cases with timestamps.

buildstockbatch/base.py
```diff
@@ -227,6 +244,8 @@ def get_clean_column_name(x):
                 return x.lower()
 
             tsdf.rename(columns=get_clean_column_name, inplace=True)
+            if cfg.get("postprocessing", {}).get("timestamps_use_period_beginning", False):
+                BuildStockBatchBase.make_period_beginning(tsdf)
```
nmerket: Might be worth turning this into a [classmethod](https://docs.python.org/3/library/functions.html#classmethod) then you can implicitly reference the class.

nmerket (commented):
Looks pretty good. One minor comment below. 

One more thought: It would be nice to have a test exercising the `make_period_beginning` method.

shorowit:
My opinion: I would still rather see this as an OS-HPXML feature than a BSB feature. Interest in this capability is not specific to ResStock, so it would be nice to make it available to a wider set of users. And it would address [this issue](https://github.com/NREL/OpenStudio-HPXML/issues/721).

shorowit:
Though I just realized that implementing it in OS-HPXML might only benefit ResStock (and other residential tools), but not ComStock.

nmerket:
All good points, @shorowit. I say we just go with this since it's already done and meets the immediate need. You can add it to OpenStudio-HPXML later to support other residential tools. We just need to make sure to not turn it on twice.

nmerket (changes requested):
This is probably ready to merge except that it needs some testing coverage of the new feature.

afontani:
I might suggest closing this and using #311

joseph-robertson:
> I might suggest closing this and using #311

If period begin timestamps are needed for ComStock, we'd probably want/need to use this PR for ComStock and #311 for ResStock.

afontani:
@joseph-robertson: Good point. 

@rHorsey , or @asparke2 : Do you want the ability to automatically convert ComStock timeseries from period ending to period beginning? There is a path for OpenStudio-HPXML to automatically do this, but @rajeee did this separately through post-processing the data. Any preferences?

asparke2:
I would strongly prefer that buildstockbatch always reports out a consistent (either beginning or ending) format, and that any users perform postprocessing after the fact to their preferred format. As soon as we put this switch in, we are going to have to remember to track which run used which setting in buildstockbatch, which will be impossible after more than a few months.  If you remember back to LA100, we had tons of problems related to how various runs were dealing with leap years and daylight savings time, which I'd really prefer not to relive.  Therefore, I would strongly prefer NOT to merge this feature.

afontani:
@asparke2 : Thanks for chiming in. For ResStock this feature is in response to OS-HPXML changing its default to period beginning timestamps (https://github.com/NREL/OpenStudio-HPXML/pull/1198), so ResStock needs some feature to keep the simulations consistent with period ending timestamps.

I have also suggested we save the YAML alongside the results in this issue to capture the exact simulation of the project (#254) which might help the remembering part of your comment.

I guess I still suggest closing this PR as ComStock will likely not use this feature.

asparke2:
It seems like instead of in buildstockbatch, the appropriate place for this code is then inside the Measure that ResStock uses to pull the timeseries results out of the E+ input file. I do like the idea of uploading the YML (and also the builstock.csv) to S3 for easier debugging for posterity though.

afontani:
@rajeee : This PR was just merged, https://github.com/NREL/buildstockbatch/pull/311. 

Are you okay with closing this PR? We might want to keep this branch for now until LA100 ES wraps up.

PR #304, opened 2022-09-02
Add ability to rerun failed array jobs

nmerket:
## Pull Request Description

Eagle is having intermittent problems where jobs will fail due to weird filesystem issues. Usually it's just a handful of the jobs. This makes it so we can just rerun those handful and not have to resubmit the whole thing. 

### Still to do

- [x] Add a "everything finished fine" message to the end of the job output and rerun only those missing that instead of ones with a "Traceback" in them. Less potentially problematic.
- [x] Have the postprocessing check that all the jobs ran successfully before gathering results and give an informative error message if not.

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -249,18 +249,18 @@ class BuildStockBatchBase(object):
 
     @staticmethod
     def validate_project(project_file):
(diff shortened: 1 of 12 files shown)


rajeee (approved):
Looks great!

Issue #327, opened 2022-11-02
Integration Tests for ComStock


nmerket:
**Is your feature request related to a problem? Please describe.**
Follow on to #318. Sometimes we make changes here that would break ComStock. 

**Describe the solution you'd like**
Having an integration test in CI would alert us to that fact before it gets merged. We're already doing this for ResStock in #326. 

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #349, opened 2023-02-16
Remove Docker dependency

### LINKED ISSUE #300: Remove Singularity dependency for Eagle and Docker Dependency for Local Runs

A lot of our complexity and bugs running on Eagle come from using Singularity to containerize openstudio. 

Now that we have official CentOS builds of openstudio, I'd like to run directly on Eagle and skip the singularity stuff. We could still automatically pull down the correct build as necessary. 

I'm not sure how this affects ComStock and their custom gem requirements. Would this make it easier or harder to work with that, @asparke2 @mpraprost ?

nmerket:
Fixes #300.

## Pull Request Description

This removes the docker dependency for local runs and the singularity dependency for runs on Eagle. Now that OpenStudio is available for more platforms and easier to get and install, the extra hassle of containerizing all the simulations is not worth it. We will still need docker for AWS because the AWS Batch service runs on Elastic Container Service, but this will simplify even that so there's only the one image that needs to be created for that.

Related to #327 since the integration tests for ComStock were tripping up becase of the custom gems. I intend to get the custom gems stuff working again in #344 after this gets merged. Having the integration tests there will ensure the code works for its intended use case. @asparke2

I would also like to see this supercede `run_analysis.rb` at some point, so we should add the features that are nice to have there to this, then we can get away with only maintaining the one thing. @joseph-robertson

## Checklist

Not all may apply

- [x] Update `LocalDockerBatch` to remove docker and change name to `LocalBatch`
- [x] Add openstudio to the ci machine
- [x] Rename stuff and clean up
    - [x] localdocker.py to local.py
    - [x] `buildstock_docker` -> `buildstock_local`
    - [x] ContainerRuntime
    - [x] sampler differences based on container or not
- [x] ~~Update `EagleBatch` to remove singularity~~
- [x] Run a small batch run on Eagle to make sure it all works if you made changes that will affect Eagle


nmerket:
I'm actually thinking of _not_ removing singularity at this time. ComStock uses OpenStudio 3.4.0 and there's not a build of that for CentOS at this time. Plus this PR is getting kind of unwieldy anyway. Thoughts @asparke2?

joseph-robertson (commented):
Using resstock to test `buildstock_docker` -> `buildstock_local` here: https://github.com/NREL/resstock/pull/1032

joseph-robertson (commented):
I tried using `weather_files_path: c:/OpenStudio/resstock/weather` (instead of `weather_files_url: https://data.nrel.gov/system/files/156/BuildStock_TMY3_FIPS.zip`) and I get `PermissionError: [Errno 13] Permission denied: 'c:\\OpenStudio\\resstock\\weather'`.

Does `weather_files_path` need to be a zip? Can we support a folder, or is this a future item to address?

joseph-robertson (approved):
Based on the resstock test branch, this appears to be working as expected. I do see just a few rounding differences in annual simulation output, but they are very small. I'll leave it up to you whether to address the weather folder issue in my previous comment.

nmerket:
> I tried using `weather_files_path: c:/OpenStudio/resstock/weather` (instead of `weather_files_url: https://data.nrel.gov/system/files/156/BuildStock_TMY3_FIPS.zip`) and I get `PermissionError: [Errno 13] Permission denied: 'c:\\OpenStudio\\resstock\\weather'`.
> 
> Does `weather_files_path` need to be a zip? Can we support a folder, or is this a future item to address?

Currently the weather files path does need to be a zip file. That's kind of a hold over from the way we do runs elsewhere. I'm open to making that change, but maybe would rather do it as another PR. That could include any other changes we want to port over from `run_analysis.rb`.

PR #350, opened 2023-02-17
Allow complex ComStock models to successfully run

rHorsey:
Our production jobs require a 5-6hr simulation per building for large complex models which are run separately to optimize AU usage.

Will write changelog etc for patch shortly.


nmerket (changes requested):
Can you put in a changelog entry for this?

rHorsey:
@nmerket  All set for you - not at all sure if I did the versioning and associated changlog updates correctly?

nmerket (approved):

PR #353, opened 2023-02-28
Skip validation from HPXMLtoOpenStudio

joseph-robertson:
Because we've changed the behavior of apply_defaults (to also apply validation) in BuildResidentialHPXML: https://github.com/NREL/OpenStudio-HPXML/pull/1298.


nmerket (changes requested):
Can you add a changelog entry for this?

nmerket (approved):

Issue #352, opened 2023-02-23
Schedules are no longer being processed into timeseries output


nmerket:
**Describe the bug**

The timeseries output isn't including the schedules even though it should. I've tracked it down to this line of code:

https://github.com/NREL/buildstockbatch/blob/acff9e6f1ec87f5f148f2c17f9de6d9e7cabccdc/buildstockbatch/base.py#L199

The schedule files have now changed and are named like `schedules20230221-10641-yoga5e.csv`, so that doesn't pick them up. When I changed that line to accept the new format for a schedule file name, I got some inconsistencies in the output parquet files on the combining step in postprocessing:

```
INFO:2023-02-23 13:19:13:buildstockbatch.postprocessing:Gathering all the parquet files in /Users/nmerket/projects/resstock/resstock/project_national/national_upgrades/parquet/timeseries/up*/*.parquet
INFO:2023-02-23 13:19:13:buildstockbatch.postprocessing:Gathered 14 files. Now writing _metadata
2023-02-23 13:19:13,814 - distributed.worker - WARNING - Compute Failed
Key:       gen-metadata-2ffd34d5b9f25259ded750abaa6a4533
Function:  aggregate_metadata
args:      ([<pyarrow._parquet.FileMetaData object at 0x11abafbf0>
  created_by: parquet-cpp-arrow version 11.0.0
  num_columns: 66
  num_rows: 2864520
  num_row_groups: 2
  format_version: 2.6
  serialized_size: 46791, <pyarrow._parquet.FileMetaData object at 0x118352ac0>
  created_by: parquet-cpp-arrow version 11.0.0
  num_columns: 66
  num_rows: 1427880
  num_row_groups: 1
  format_version: 2.6
  serialized_size: 45075, <pyarrow._parquet.FileMetaData object at 0x11abae610>
  created_by: parquet-cpp-arrow version 11.0.0
  num_columns: 66
  num_rows: 1427880
  num_row_groups: 1
  format_version: 2.6
  serialized_size: 46617, <pyarrow._parquet.FileMetaData object at 0x11a54f240>
  created_by: parquet-cpp-arrow version 11.0.0
  num_columns: 66
  num_rows: 1419120
  num_row_groups: 1
  format_version: 2.6
  serialized_size: 45034, <pyarrow._parquet.FileMetaData object at 0x11abaf420>
  created_by: parquet-cpp-arrow version 11.0.0
  num_columns: 66
  num_rows: 1410360
  num_row_groups: 1
[...]

PR #355, opened 2023-03-02
Force schedule columns to float

rajeee:
Addresses #352

Sometimes a building's particular schedule column can have integer only values which results in pandas assigning integer datatype to that column. But this causes problem when generating the metadata during postprocessing since other schedule file can have float datatypes. This PR fixes the issue by always interpreting the schedule values as float during read.


nmerket (changes requested):
Can you put in a changelog entry for this?

nmerket (approved):

PR #363, opened 2023-04-03
ResStock-HPXML: include_annual arguments

joseph-robertson:
Support the following ReportSimulationOutput changes:
* A bunch of `include_annual_foo` are now available. We will fix (i.e., user has no option to override) requesting all annual outputs, except for system use consumptions which will always not be requested.
* Also fixes new `include_timeseries_system_use_consumptions` argument to always false.


nmerket (changes requested):
Just pulling up the checklist from the pull request template...

- [ ] Tests exercising your feature/bug fix (check coverage report on Checks -> BuildStockBatch Tests -> Artifacts) _Is it worth adding a simple test around this feature in the test_workflow_generator.py?_
- [ ] Run a small batch run on Eagle to make sure it all works if you made changes that will affect Eagle

joseph-robertson:
New `test_workflow_generator.py` checks are commented out because the test checks against resstock develop... but the new arguments aren't on develop yet...

nmerket (commented):
I think this is good. At what point do we merge this? Before or after the corresponding resstock change?

joseph-robertson:
I think we merge here, then release buildstockbatch. Then I can go into the resstock PR and point to the tagged buildstockbatch release before merging it into develop. Does that sound right?

nmerket (approved):

PR #365, opened 2023-04-19
upload-buildstockcsv-to-s3

### LINKED ISSUE #348: Upload buildstock.csv to S3 during postprocessing
Upload buildstock.csv to S3 during postprocessing.

Upload buildstock.csv to S3 during postprocessing. This is particularly useful for ComStock since the sample is always precomputed. This makes it easier to postprocess the results. Could also consider uploading the YML file for easier correspondence between runs on Eagle and S3 directories.

yingli-NREL:
Modified postprocessing.py def upload_results() to upload buildstockcsv to s3

## Checklist

Not all may apply

- [x] Run a small batch run on Eagle to make sure it all works if you made changes that will affect Eagle


nmerket:
@yingli-NREL I added the pull request checklist back in. What you need to do is make sure you've done each item in that list (or verified that it wasn't applicable). If you're not sure what some of those mean, @rajee or I can help.

yingli-NREL:
create one folder named "buildstock_csv" in the S3 results directory and add buildstock.csv in the folder "buildstock_csv"

buildstockbatch/postprocessing.py
```diff
@@ -628,8 +632,17 @@ def upload_file(filepath):
...
         s3key = Path(s3_prefix_output).joinpath(filepath).as_posix()
         bucket.upload_file(str(full_path), str(s3key))
+    
+    def upload_buildstock_csv(filepath):
+        full_path = buildstock_dir.joinpath(filepath)
+        s3 = boto3.resource('s3')
+        bucket = s3.Bucket(s3_bucket)
+        s3_prefix_output_new = s3_prefix_output+ '/' + 'buildstock_csv' + '/'
+        s3key = Path(s3_prefix_output_new).joinpath(filepath).as_posix()
+        bucket.upload_file(str(full_path), str(s3key))
```
nmerket: When I look at the coverage report, this doesn't seem to be getting called. It's not clear to me why.

image

It would make sense to add something to [this test](https://github.com/NREL/buildstockbatch/blob/348-upload-buildstockcsv-to-s3-during-postprocessing/buildstockbatch/test/test_base.py#L236) to check if the file got uploaded. It's using [mock](https://docs.python.org/3/library/unittest.mock.html) to not actually upload file to s3 but to see that it was called successfully.

buildstockbatch/postprocessing.py
```diff
@@ -607,6 +608,9 @@ def upload_results(aws_conf, output_dir, results_dir):
         all_files.append(file.relative_to(parquet_dir))
     for file in [*ts_dir.glob('_common_metadata'), *ts_dir.glob('_metadata')]:
         all_files.append(file.relative_to(parquet_dir))
+    buildstock_csv = []   
+    for file in buildstock_dir.glob('buildstock.csv'):
+        buildstock_csv.append(file.relative_to(buildstock_dir))
```
nmerket: I see why you made this a list so you could use `map` below, but that's not necessary. You can just pull the filename here. We should also only upload it if it exists. I think ComStock might put this file somewhere different, so it may not always be where you're looking for it.

```suggestion
    buildstock_csv = buildstock_dir / 'buildstock.csv'
```

I haven't tested that this works. I might be missing some normalization or something.

buildstockbatch/postprocessing.py
```diff
@@ -628,8 +632,17 @@ def upload_file(filepath):
...
+    def upload_buildstock_csv(filepath):
+        full_path = buildstock_dir.joinpath(filepath)
+        s3 = boto3.resource('s3')
+        bucket = s3.Bucket(s3_bucket)
+        s3_prefix_output_new = s3_prefix_output+ '/' + 'buildstock_csv' + '/'
+        s3key = Path(s3_prefix_output_new).joinpath(filepath).as_posix()
+        bucket.upload_file(str(full_path), str(s3key))
 
     dask.compute(map(dask.delayed(upload_file), all_files))
+    dask.compute(map(dask.delayed(upload_buildstock_csv), buildstock_csv))
```
nmerket: It's best to gather all the work you want to do into one list of tasks and then send that to `dask.compute` once. You don't have to call `dask.delayed` with a `map`, it was just convenient in this case. See [dask.delayed documentation](https://docs.dask.org/en/stable/delayed.html) for more details. This change depends on `buildstock_csv` being a filename, not a list with just one item in it. 

```suggestion
    tasks = list(map(dask.delayed(upload_file), all_files))
    tasks.append(dask.delayed(upload_buildstock_csv)(buildstock_csv))
    dask.compute(tasks)
```

buildstockbatch/postprocessing.py
nmerket: One more thought on this, could we not have `upload_buildstock_csv` here? It seems to mostly be a copy of the more generic `upload_file` but puts the file somewhere else. Maybe there could be a second, optional argument to `upload_file` where you could specify the s3 location when it's different than the default and then just use it for buildstock.csv.

nmerket (changes requested):
I apologize for taking so long to properly review this. Have you run a small batch to verify this is working? This could use some unit testing, which is kind of tricky with s3 stuff because you'll probably need to mock things. There's some existing tests I linked to that could be added to. A couple other notes below. I know it is a lot, but I've been thorough and somewhat critical hopefully as a teaching tool. If it would help, we can meet to talk through some of this.

buildstockbatch/postprocessing.py
yingli-NREL: I created a local environment in my HPC account and used that environment to run a small batch. It works. The results are upload in Amazon S3>Buckets>eulp>buildstock_csv_to_s3/>test/>test10/

nmerket:
The buildstock.csv is in a different place depending on whether you're running `buildstock_eagle` or `buildstock_local`. When I run `buildstock_local` and have the output_directory not in the default location, it fails to upload:

```
INFO:2023-10-06 17:06:19:buildstockbatch.postprocessing:Uploading the parquet files to s3
2023-10-06 17:06:20,315 - distributed.worker - WARNING - Compute Failed
Key:       upload_buildstock_csv-3a9eafb5-350b-4d0d-ad8d-efae94791c46
Function:  upload_buildstock_csv
args:      ('')
kwargs:    {}
Exception: "FileNotFoundError(2, 'No such file or directory')"
```

I can take a look at why next week.

buildstockbatch/postprocessing.py
```diff
@@ -595,6 +595,7 @@ def remove_intermediate_files(fs, results_dir, keep_individual_timeseries=False)
 def upload_results(aws_conf, output_dir, results_dir):
     logger.info("Uploading the parquet files to s3")
 
+    buildstock_dir = Path(results_dir).parent.joinpath('housing_characteristics')
```
nmerket: I think this is the problem right here. It's assuming that file is in the usual place it is on Eagle. I think the sampler returns the buildstock.csv location. We should catch that and use that file. Could be a bit tricky because the sample is called way before the postprocessing on a different node and everything.

buildstockbatch/postprocessing.py
```diff
@@ -607,6 +608,9 @@ def upload_results(aws_conf, output_dir, results_dir):
         all_files.append(file.relative_to(parquet_dir))
     for file in [*ts_dir.glob('_common_metadata'), *ts_dir.glob('_metadata')]:
         all_files.append(file.relative_to(parquet_dir))
+    buildstock_csv = ''
+    for file in buildstock_dir.glob('buildstock.csv'):
+        buildstock_csv = file.relative_to(buildstock_dir)
```
nmerket: Also, this is a little bizarre to use a for loop and glob to look for one file.

nmerket (commented):
A few more notes. I'll try to get this working early next week.

nmerket (approved):
I confirmed it's working for both the local version and on Eagle.

PR #366, opened 2023-05-04
Add references section to yaml schema

rajeee:
Add a throwaway section in the yaml schema called references where yaml anchor references can be defined to be used in other places. Without this dedicated section, there is no way to define all the references at single location and one would have to define their anchor at the first point of use.


nmerket:
Can you add a changelog entry and maybe a test that shows this working?

nmerket (approved):

PR #351, opened 2023-02-22
ResStock-HPXML: simple bill calcs (enhancement)

joseph-robertson:
See https://github.com/NREL/resstock/pull/1012 for context.

PR #374, opened 2023-05-26
Exclude None from default_na_values list when reading csv files

### LINKED ISSUE #373: residential_quota_downselect leads to bad format buildstock
**Describe the bug**
There's a problem with the buildstock created from`residential_quota_downselect`. All "None" fields become empty and I believe there are other formatting issues that are causing these failures on Eagle:

```
[00:32:57.683956 ERROR] Could not find parameter 'Geometry Building Number Units MF' and option '6.0' in /lib/resources/options_lookup.tsv.

[00:16:10.751502 ERROR] Could not find parameter 'CEC Climate Zone' and option '13.0' in /lib/resources/options_lookup.tsv.

[22:32:00.491794 ERROR] :/ruby/2.7.0/gems/openstudio-workflow-2.3.1/lib/openstudio/workflow/util/measure.rb failed with message Runner error :/ruby/2.7.0/gems/openstudio-workflow-2.3.1/lib/openstudio/workflow/util/measure.rb failed with invalid null reference Expected argument 2 of type std::string const &, but got NilClass nil
```
The last error I think is tied to None being completely gone.

**To Reproduce**
This folder contains the yaml, downselect buildstock and original buildstock. You can run the yaml against ResStock develop branch.

I am using bsb env: `buildstock-2023.05.0`
@nmerket for your awareness. This is potentially caused by recent pandas changes and it does not appear to be an issue for `buildstock-2023.01.0`

[downselect buildstock errors.zip](https://github.com/NREL/buildstockbatch/files/11566322/downselect.buildstock.errors.zip)

rajeee:
[Pandas 2.0 added "None" to default na_values in read_csv()](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#other-api-changes).

This causes breaking change from previous version when reading buildstock.csv or other files that has the string "None". This PR adds a custom read_csv function that restores previous behavior. It should also fix #373.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,8 +34,8 @@ jobs:
           wget --quiet https://data.nrel.gov/system/files/156/BuildStock_TMY3_FIPS.zip
       - name: Download and Install OpenStudio
         run: |
-          wget -q https://github.com/NREL/OpenStudio/releases/download/v3.5.1/OpenStudio-3.5.1+22e1db7be5-Ubuntu-20.04.deb
-          sudo apt install -y ./OpenStudio-3.5.1+22e1db7be5-Ubuntu-20.04.deb
+          wget -q https://github.com/NREL/OpenStudio/releases/download/v3.6.1/OpenStudio-3.6.1+bb9481519e-Ubuntu-20.04-x86_64.deb
+          sudo apt install -y ./OpenStudio-3.6.1+bb9481519e-Ubuntu-20.04-x86_64.deb
           openstudio openstudio_version
           which openstudio
       - name: Install buildstockbatch
--- a/buildstockbatch/base.py
+++ b/buildstockbatch/base.py
@@ -314,7 +314,38 @@ class BuildStockBatchBase(object):
         except AttributeError:
             raise ValidationError(f'Sampler class `{sampler_name}` is not available.')
         args = cfg['sampler']['args']
-        return Sampler.validate_args(project_file, **args)
+        Sampler.validate_args(project_file, **args)
+        if issubclass(Sampler, sampler.PrecomputedSampler):
+            sample_file = cfg['sampler']['args']['sample_file']
+            if not os.path.isabs(sample_file):
+                sample_file = os.path.join(os.path.dirname(project_file), sample_file)
+            else:
+                sample_file = os.path.abspath(sample_file)
+            buildstock_df = read_csv(sample_file)
+            BuildStockBatchBase.validate_buildstock_csv(project_file, buildstock_df)
+        return True
+
+
+    @staticmethod
+    def validate_buildstock_csv(project_file, buildstock_df):
+        param_option_dict, _ = BuildStockBatchBase.get_param_option_dict(project_file)
+        # verify that all the Columns in buildstock_df only have values available in param_option_dict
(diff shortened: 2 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -34,8 +34,8 @@ jobs:
           wget --quiet https://data.nrel.gov/system/files/156/BuildStock_TMY3_FIPS.zip
       - name: Download and Install OpenStudio
         run: |
-          wget -q https://github.com/NREL/OpenStudio/releases/download/v3.5.1/OpenStudio-3.5.1+22e1db7be5-Ubuntu-20.04.deb
-          sudo apt install -y ./OpenStudio-3.5.1+22e1db7be5-Ubuntu-20.04.deb
+          wget -q https://github.com/NREL/OpenStudio/releases/download/v3.6.1/OpenStudio-3.6.1+bb9481519e-Ubuntu-20.04-x86_64.deb
+          sudo apt install -y ./OpenStudio-3.6.1+bb9481519e-Ubuntu-20.04-x86_64.deb
```
nmerket, line 38: Is the new version of OpenStudio required for this? It looks like @joseph-robertson has updated it in #351 as well.

buildstockbatch/base.py
```diff
@@ -315,7 +314,38 @@ def validate_sampler(project_file):
...
+                sample_file = os.path.join(os.path.dirname(project_file), sample_file)
+            else:
+                sample_file = os.path.abspath(sample_file)
+            buildstock_df = read_csv(sample_file, dtype=str)
+            BuildStockBatchBase.validate_buildstock_csv(project_file, buildstock_df)
+        return True
+
+
+    @staticmethod
+    def validate_buildstock_csv(project_file, buildstock_df):
```
nmerket, line 330: What's this doing? Tell me in the changelog.

nmerket (changes requested):
Looks pretty good. Well done on the testing. Can you add an entry to the changelog?

Also, could you require `pandas>2` in setup.py for this? It probably won't work with pandas 1.x anymore.

.github/workflows/ci.yml
rajeee: Yeah, the develop was missing this update to CI so I did it as a part of this PR. Between Joe's and this, whichever PR gets merged last will have to update from develop (or it might be fine since they are exact same changes).

buildstockbatch/base.py
rajeee: Done.

rajeee:
> Looks pretty good. Well done on the testing. Can you add an entry to the changelog?
> 
> Also, could you require `pandas>2` in setup.py for this? It probably won't work with pandas 1.x anymore.

Thanks. It should still work for pandas 1.x.

setup.py
```diff
@@ -33,7 +33,7 @@
         'pyyaml',
         'requests',
         'numpy',
-        'pandas',
+        'pandas>2.0',
```
nmerket:
```suggestion
        'pandas>=2',
```

nmerket (commented):
Should let it be equal to as well.

nmerket (approved):

afontani:
Thanks, @nmerket and @rajeee!

PR #396, opened 2023-10-10
Allows non-integer minutes_per_sim

### LINKED ISSUE #377: Allow minutes_per_sim to be less than a minute.
ResStock simulation time has improved a lot over time and based on some empirical data in Eagle test runs, it is now about 1/3 mintues. However, `minutes_per_sim` only takes integer value. https://github.com/NREL/buildstockbatch/blob/f37754c3e2bd85af7a687419925745e271e75ba6/buildstockbatch/schemas/v0.3.yaml#L42

Change the argument from minutes_per_sim to seconds_per_sim to go below minute level.

**Describe alternatives you've considered**
We can just use 1 minute and just deal with getting larger walltime than necessary. However, this will cause unnecessary scheduling wait.

**Additional context**

nmerket:
Fixes #377.

I tested it on Eagle and it has the intended effect.

### DIFF
--- a/buildstockbatch/schemas/v0.3.yaml
+++ b/buildstockbatch/schemas/v0.3.yaml
@@ -39,7 +39,7 @@ aws-emr-spec:
 
 hpc-spec:
   account: str(required=True)
-  minutes_per_sim: int(max=480, required=False)
+  minutes_per_sim: num(min=0.05, max=480, required=True)
   n_jobs: int(required=False)
   postprocessing: include('hpc-postprocessing-spec', required=False)
   sampling: include('sampling-spec', required=False)
--- a/buildstockbatch/test/test_eagle.py
+++ b/buildstockbatch/test/test_eagle.py
@@ -3,7 +3,6 @@ import json
 import os
 import pandas as pd
 import pathlib
-import requests
 import shutil
 import tarfile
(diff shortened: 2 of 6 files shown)


buildstockbatch/schemas/v0.3.yaml
```diff
@@ -39,7 +39,7 @@ aws-emr-spec:
 
 hpc-spec:
   account: str(required=True)
-  minutes_per_sim: int(max=480, required=False)
+  minutes_per_sim: num(min=0.05, max=480, required=True)
```
afontani, line 42: Looks like we will need another 10x reduction before this becomes an issue. Probably safe.

buildstockbatch/test/test_eagle.py
```diff
@@ -195,6 +194,33 @@ def test_qos_high_job_submit(mock_subprocess, basic_residential_project_file, mo
         assert '--qos=high' in mock_subprocess.run.call_args[0][0]
 
 
+def test_queue_jobs_minutes_per_sim(mocker, basic_residential_project_file, monkeypatch):
```
afontani, line 197: I am glad to see this test.

afontani (approved):

PR #399, opened 2023-10-12
remove always-added reporting measure from Comstock

eringold:
SimulationOutputReport was being added to the comstock workflow by default, **as well as** when specified in the yml. This removes the default addition: SimulationOutputReport must be specified in the yml to be included in the workflow. 

Eyeballing from my recent profiling of a 10k run shows the second instance of SimulationOutputReport taking roughly 10% of the total workflow runtime.


asparke2 (approved):

nmerket (commented):
Looks pretty good. Could you put something in the changelog for this?

eringold:
Thanks @nmerket, I've updated the changelog.

nmerket (approved):

PR #400, opened 2023-10-13
Add config option for maximum simulation time for individual models

asparke2:
ComStock has some very long-running models which often have to be manually killed on Eagle, otherwise the entire job will time out on SLURM. When the job times out, the entire job must be re-run. This feature adds a config option to set a max simulation time that's enforced for each individual model. When this max time is reached, the openstudio process for this model is terminated and the relevant log files are generated such that postprocessing can handle these jobs correctly.


buildstockbatch/eagle.py
```diff
@@ -298,6 +298,7 @@ def run_building_d(i, upgrade_idx):
             json.dump(dpouts, f)
 
         # Compress simulation results
+        time.sleep(60)  # Allow results JSON to finish writing
```
nmerket: I don't love this. What's going on that we have to wait here? Could we do some sort of check to make sure it exists in a little loop?

buildstockbatch/eagle.py
```diff
@@ -390,6 +391,9 @@ def run_building(cls, output_dir, cfg, n_datapoints, i, upgrade_idx=None):
                 env_vars = dict(os.environ)
                 env_vars['SINGULARITYENV_BUILDSTOCKBATCH_VERSION'] = bsb_version
                 logger.debug('\n'.join(map(str, args)))
+                max_time_min = cfg.get('max_minutes_per_sim', 525600)
```
nmerket: 🎵 525,600 minutes 🎵 

This is definitely functionally an "infinite" timeout, but I think I'd prefer to not pass a timeout if there isn't one in the yaml file. Something like 

```python
max_time_min = cfg.get('max_minutes_per_sim'). # returns None if key not present
if max_time_min is not None:
    subprocess_kw = {'timeout': max_time_s}
else:
    subprocess_kw = {}
...
subprocess.run(
    args,
    ...,
    **subprocess_kw
)
```

buildstockbatch/eagle.py
```diff
@@ -400,7 +404,26 @@ def run_building(cls, output_dir, cfg, n_datapoints, i, upgrade_idx=None):
...
+                                'completed_at': end_time.strftime('%Y%m%dT%H%M%SZ'),
+                                'completed_status': 'Fail',
+                                'timeout': msg
+                            }
+                            out_osw.write(json.dumps(out_msg, indent=3))
+                        with open(os.path.join(sim_dir, 'run', 'out.osw'), 'a') as run_log:
+                            run_log.write(f"[{end_time.strftime('%H:%M:%S')} ERROR] {msg}")
+                        with open(os.path.join(sim_dir, 'run', 'failed.job'), 'w') as failed_job:
+                            failed_job.write(f"[{end_time.strftime('%H:%M:%S')} ERROR] {msg}")
+                        time.sleep(60)  # Wait for EnergyPlus to release file locks and data_point.zip to finish
```
nmerket, line 430: Again this feels clunky. Why hasn't this been a problem up until now?

buildstockbatch/eagle.py
```diff
@@ -390,6 +391,9 @@ def run_building(cls, output_dir, cfg, n_datapoints, i, upgrade_idx=None):
                 env_vars = dict(os.environ)
                 env_vars['SINGULARITYENV_BUILDSTOCKBATCH_VERSION'] = bsb_version
                 logger.debug('\n'.join(map(str, args)))
+                max_time_min = cfg.get('max_minutes_per_sim', 525600)
+                max_time_s = max_time_min * 60
+                start_time = dt.datetime.now()
                 with open(os.path.join(sim_dir, 'singularity_output.log'), 'w') as f_out:
```
nmerket: I keep meaning to do this, but let's change this filename to "openstudio_output.log" for consistency across implementations.

buildstockbatch/local.py
```diff
@@ -183,36 +185,60 @@ def run_building(cls, buildstock_dir, weather_dir, results_dir, measures_only,
...
+            proc_output = subprocess.run(
+                run_cmd,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                env=env_vars,
+                cwd=sim_dir,
+                timeout=max_time_s
+            )
+            with open(sim_path / 'openstudio_output.log', 'wb') as f_out:
+                f_out.write(proc_output.stdout)
```
nmerket: I think we should change this to write out to a file more like eagle.py does. That way we get the openstudio output even if there's a timeout. This way we won't.

buildstockbatch/local.py
```diff
@@ -257,6 +283,7 @@ def run_batch(self, n_jobs=None, measures_only=False, sampling_only=False):
             n_jobs = -1
         dpouts = Parallel(n_jobs=n_jobs, verbose=10)(all_sims)
 
+        time.sleep(10)
```
nmerket, line 291: I'm really interested to hear the story behind all these `time.sleep`s.

nmerket (changes requested):
A few notes below. I also noticed that the ResStock integration tests are failing, so we'll need to figure that out before that can merge. I'll take a look and see if I can see why.

nmerket (approved):

PR #407, opened 2023-11-01
ResStock-HPXML: simulating whole SFA/MF buildings

joseph-robertson:


joseph-robertson:
OS-HPXML was updated to replace the building_id "ALL" enum with a bool that lives directly in the HPXML file.

PR #409, opened 2023-11-03
Small refactor to simplify adding a GCP implementation

nweires:
Move a few things out of aws.py that will be reused in a GCP implementation, to keep future changes simpler.


buildstockbatch/docker_base.py
lathanh, line 1: How about putting this in a subdir like 'cloud' or 'docker' and also putting the Dockerfile in there (cleaning up the root dir a bit), too? Especially since we'd like to refactor more out of aws.py, likely into yet another file (e.g., cloud_base.py)

lathanh (approved):

buildstockbatch/docker_base.py
nmerket: Sounds good to me.

nmerket:
It seems our CI is failing because it's expecting PRs from _within_ this repo and some of the actions don't have the correct permissions. Don't worry about that too much for now. I'll see if I can sort out a way to make that work before merging it down.

buildstockbatch/docker_base.py
nweires: Okay, moved docker_base.py in a new cloud/ directory.

@lathanh I'll let you move the Dockerfile separately, since you worked on that piece and I think docker can be finicky about file locations.

nweires:
@nmerket Are you okay with merging this now? Or do you want to work on getting the checks to pass first?

nmerket:
Yeah, I can merge this. It looks like the issues we're seeing is a [known issue](https://github.com/5monkeys/cobertura-action#cobertura-action) with PRs from a fork for one of the actions we're using. It's not that important and it looks like all tests are passing otherwise. However, If you have any suggestions on how to make github happy with this, I'm all ears.

PR #410, opened 2023-11-03
Update cost multiplier link in upgrade scenarios docs

### LINKED ISSUE #408: Update cost multiplier link in upgrade scenarios docs
The cost multiplier link [here](https://github.com/NREL/buildstockbatch/blob/661896200a3589c107f502bbc020a509e68de75b/docs/project_defn.rst#upgrade-scenarios) doesn't go to the right place anymore. 

We should also have a link to the comstock list

- [ResStock list](https://github.com/NREL/resstock/blob/main/measures/ApplyUpgrade/resources/constants.rb)
- [ComStock list](https://github.com/NREL/ComStock/blob/main/measures/ApplyUpgrade/measure.rb#L77)

joseph-robertson:
Closes #408.


docs/project_defn.rst
```diff
@@ -122,8 +122,10 @@ following properties:
           Since there can be multiple costs, this permits both fixed and variable costs for upgrades
           that depend on the properties of the baseline building.
           The multiplier needs to be from
-          `this enumeration list inthe resstock or comstock repo <https://github.com/NREL/resstock/blob/master/measures/ApplyUpgrade/measure.rb#L71-L87>`_
-          or from the list in your branch of that repo.
+          this enumeration list in the
+          `resstock <https://github.com/NREL/resstock/blob/main/measures/ApplyUpgrade/resources/constants.rb#L12-L38>`_ or
```
joseph-robertson: Alternatively, it may be appropriate to point to resstock's RTD that displays cost multiplier types: https://resstock.readthedocs.io/en/latest/workflow_outputs/default_outputs.html#upgrade-costs

nmerket: Works for me. Can you make that change?

nmerket: Wait a minute. But those aren't the enumerations that one would put in the input file. That's the context here. If someone tries to put in `upgrade_costs.door_area_ft_2` instead of `Door Area (ft^2)` it won't work.

joseph-robertson: Good point. The RTD should probably include the exact enums. But out of scope here. Let's just stick with pointing to `constants.rb`.

nmerket (changes requested):
It looks like the docs build is failing on CI. Can you look into that?

nmerket (approved):

PR #416, opened 2023-11-15
Fix CI workflow for PRs from other forks

nweires:
The CI workflow currently fails for pull requests from other forks because it only has read access to the repo (for good security reasons), which means it can't post comments/edits/updates to the PR.

This splits things into two separate workflows:
- CI workflow: Read-only because it executes code from the pull request, which could potentially be malicious. Runs tests and saves code coverage results.
- Coverage workflow: Has write access, but only the version on the default (`develop`) branch is run, and it doesn't execute any code from outside PRs. Reads the code coverage results and posts them as a comment on the PR.

See the links in the comments for more details about this setup.

I tested this on the rewiringamerica/buildstockbatch fork and confirmed that the coverage results were posted on PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,14 +67,12 @@ jobs:
         with:
           name: coverage-report-html
           path: buildstockbatch/coverage/htmlreport/
-      - name: Report coverage to PR
-        uses: 5monkeys/cobertura-action@v13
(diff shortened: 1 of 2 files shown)


lathanh (approved):
a bit too Greek to me, but LGTM for our end.

nmerket (approved):
Thank you for fixing this!

PR #417, opened 2023-11-17
Skip post-processing of upgrade directories that fail to list

lathanh:
### Background
* Upgrade directories may be empty if the upgrade is invalid.
* In some cloud filesystems, there aren't actual directories (including for [S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-folders.html) and [Google Cloud Storage](https://cloud.google.com/storage/docs/folders)). So, an "empty directory" means the directory doesn't exist and can't be listed.

### Problem
Trying to list an "empty" upgrade directory during post-processing (which can happen if the upgrade was invalid) fails with an exception for Google Cloud Storage.

### Fix
Treat a failure to list an upgrade directory during post-processing the same as when the directory is empty (i.e., skip it).

### Alternatives considered

1. **Create a fake directory**: This causes the directory listing to also include the fake directory itself, causing a different error (trying to read the directory as if it's a parquet file), and thus a change to also list only parquet files in the directory (i.e., requires a hack to handle the hack).

## Checklist

Not all may apply

- [x] ~~Tests exercising your feature/bug fix (check coverage report on Checks -> BuildStockBatch Tests -> Artifacts)~~
- [x] ~~Run a small batch run on Eagle to make sure it all works if you made changes that will affect Eagle~~
- [x] ~~Add to the changelog_dev.rst file and propose migration text in the pull request~~

### DIFF
--- a/buildstockbatch/postprocessing.py
+++ b/buildstockbatch/postprocessing.py
@@ -514,7 +514,14 @@ def combine_results(fs, results_dir, cfg, do_timeseries=True):
         if do_timeseries:
             # Get the names of the timeseries file for each simulation in this upgrade
             ts_upgrade_path = f"{ts_in_dir}/up{upgrade_id:02d}"
-            ts_filenames = [ts_upgrade_path + ts_filename for ts_filename in fs.ls(ts_upgrade_path)]
+            try:
+                ts_filenames = [ts_upgrade_path + ts_filename for ts_filename in fs.ls(ts_upgrade_path)]
+            except:
+                # Upgrade directories may be empty if the upgrade is invalid. In some cloud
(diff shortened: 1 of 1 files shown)


nmerket (commented):
Seems pretty straightforward to me. Like in the other PR, I added the checklist from our PR template (which should be working right again). It helps me to make sure we've sorted all things that can come up. If something doesn't apply to this PR, just ~~cross it out~~ and check the box.

lathanh:
> Like in the other PR, I added the checklist from our PR template (which should be working right again). It helps me to make sure we've sorted all things that can come up. If something doesn't apply to this PR, just ~cross it out~ and check the box.

Done! Let me know if there's anything else.

nmerket (approved):

PR #423, opened 2023-12-08
Add GCP implementation

nweires:
## Pull Request Description

Adds a GCP implementation of Buildstock Batch. This includes:
- `gcp/README.md` - Overview of the architecture of the job on GCP.
- `gcp/gcp.py` - The bulk of the implementation code. It roughly follows the structure of the AWS implementation. Where possible, shared logic was already moved into `docker_base.py` to avoid duplicated code.
- `docs/` - Documentation for installation and setup, configuration options, and running jobs.
- `schemas/v0.3.yaml` - Schema for the new GCP-related project configuration options.

I recommend doing a "Squash and Merge" of this PR - we don't need the (many!) individual commits that it includes.

## Checklist

Not all may apply

  - Added exception for the `gcp/` directory, but shared code in docker_base.py has tests.


setup.py
```diff
@@ -56,17 +71,22 @@
...
-            "black",
+            "black~=24.0",
             "rope",
             "doc8",
             "pre-commit",
         ]
+        + gcp_requires
+        + aws_requires,
+        "gcp": gcp_requires,
+        "aws": aws_requires,
```
nmerket, line 82: This is clever. I like that I can install `.[dev]` and get both aws and gcp.

docs/installation.rst
```diff
@@ -64,6 +64,9 @@ For Windows, the process is similar.
 .. _set a Windows environment Variable: https://www.computerhope.com/issues/ch000549.htm
 .. _OpenStudio release: https://github.com/NREL/OpenStudio/releases
 
+.. _python:
+
+.. _bsb-python:
```
nmerket: I'm a little confused why both of these are needed.

buildstockbatch/gcp/gcp.py
```diff
@@ -0,0 +1,1267 @@
...
+
+class GcpBatch(DockerBatchBase):
+    # https://patorjk.com/software/taag/#p=display&f=Santa%20Clara&t=BuildStockBatch%20%20%2F%20GCP
+    LOGO = """
+     _ __         _     __,              _ __                      /     ,___ ,___ _ __
+    ( /  )    o  //   /(    _/_       / ( /  )     _/_    /       /     /   //   /( /  )
+     /--< , ,,  // __/  `.  /  __ _, /<  /--< __,  /  _, /_      /     /  __/      /--'
+    /___/(_/_(_(/_(_/_(___)(__(_)(__/ |_/___/(_/(_(__(__/ /_    /     (___/(___/  /
+      Executing BuildStock projects with grace since 2018
+"""
```
nmerket, line 160: Nice ascii art update.

buildstockbatch/gcp/gcp.py
```diff
@@ -0,0 +1,1267 @@
...
+        return f"projects/{self.gcp_project}/locations/{self.region}/jobs/{self.postprocessing_job_id}"
+
+    @property
+    def postprocessing_job_console_url(self):
+        return (
+            f"https://console.cloud.google.com/run/jobs/details/{self.region}"
+            f"/{self.postprocessing_job_id}/executions?project={self.gcp_project}"
+        )
+
+    # todo: aws-shared (see file comment)
```
nmerket: Agree that it would be good to consolidate this with the AWS build_image, but I do get that different things could be needed.

buildstockbatch/gcp/gcp.py
```diff
@@ -0,0 +1,1267 @@
...
+        gcp_cfg = self.cfg["gcp"]
+        job_env_cfg = gcp_cfg.get("job_environment", {})
+        resources = batch_v1.ComputeResource(
+            cpu_milli=1000 * job_env_cfg.get("vcpus", 1),
+            memory_mib=job_env_cfg.get("memory_mib", 1024),
+            boot_disk_mib=job_env_cfg.get("boot_disk_mib", None),
+        )
+
+        # Give three minutes per simulation, plus ten minutes for job overhead
+        task_duration_secs = 60 * (10 + batch_info.n_sims_per_job * 3)
```
nmerket: This should be configurable in the project yaml file. Comstock simulations take on average of 10 minutes each. See the `hpc-spec` where we have to estimate a wall time submitting via sbatch.

buildstockbatch/gcp/gcp.py
```diff
@@ -0,0 +1,1267 @@
...
+                logger.debug("Downloading {}.gz".format(epw_filename))
+                bucket.blob(f"{gcs_prefix}/weather/{epw_filename}.gz").download_to_file(f_gz)
+                with open(weather_dir / epw_filename, "wb") as f_out:
+                    logger.debug("Extracting {}".format(epw_filename))
+                    f_out.write(gzip.decompress(f_gz.getvalue()))
+
+        cls.run_simulations(cfg, task_index, jobs_d, sim_dir, GCSFileSystem(), f"{gcs_bucket}/{gcs_prefix}")
+
+    # todo: Do cleanup (which the aws script does, in the 'nrel/aws_batch' branch)
+    # todo: aws-shared (see file comment): Such cleanup should be shared with the aws script.
```
nmerket: What follow up should we do here?

nmerket (commented):
Looks pretty good. Made some comments below. A few other notes:

- We'll need to bring this up to date with `develop` probably after the other PRs merge in.
- On AWS and HPC we put the data in S3 and then run a data crawler on it so we can run sql queries on it in AWS Athena (Trino DB). I noticed you're not doing an equivalent thing for GCP here. Is that because there's no equivalent service or it just wasn't necessary for your use case?

Thanks for all your contributions here. Let me know how I can help get this across the finish line.

docs/installation.rst
nweires: I'm guessing this was leftover from a merge conflict or something. Cleaning it up now!

buildstockbatch/gcp/gcp.py
nweires: Add a config option for this.

buildstockbatch/gcp/gcp.py
nweires: Moved in #443.

buildstockbatch/gcp/gcp.py
nweires: Cleaned this up.

buildstockbatch/postprocessing.py
```diff
@@ -18,6 +18,7 @@
 from dask.dataframe.io.parquet import create_metadata_file
 from fsspec.implementations.local import LocalFileSystem
 from functools import partial
+from gcsfs import GCSFileSystem
```
nmerket: postprocessing.py is used by all our implementations. `gcsfs` is only installed if the `gcp` extras are installed. This will cause import errors for all other implementations where they may not have installed those extras. It looks like none of these imported fsspec filesystem classes are even used in this file. We can just remove the imports.

```suggestion
```

buildstockbatch/postprocessing.py
```diff
@@ -18,6 +18,7 @@
 from dask.dataframe.io.parquet import create_metadata_file
 from fsspec.implementations.local import LocalFileSystem
```
nmerket:
```suggestion
```

buildstockbatch/postprocessing.py
```diff
@@ -32,6 +33,7 @@
 from s3fs import S3FileSystem
```
nmerket:
```suggestion
```

docs/installation.rst
```diff
@@ -141,7 +141,7 @@ You can get a list of installed environments by looking in the envs directory
 
    ls /kfs2/shared-projects/buildstock/envs
 
-Developer Installaion
+Developer Installation
```
nmerket, line 144: Thanks for fixing my typo.

docs/installation.rst
```diff
@@ -245,7 +245,7 @@ Install either `Docker Desktop <https://docs.docker.com/get-docker/>`_ of
 BuildStockBatch Python Library
 ..............................
 
-Instal the buildstockbatch python library as described in :ref:`bsb-python` for
+Install the buildstockbatch python library as described in :ref:`bsb-python` for
```
nmerket, line 248: Spelling is hard.

nmerket (changes requested):
This looks really good. One note below on imports.

buildstockbatch/postprocessing.py
nweires: Good call - fixed!

nweires:
> * On AWS and HPC we put the data in S3 and then run a data crawler on it so we can run sql queries on it in AWS Athena (Trino DB). I noticed you're not doing an equivalent thing for GCP here. Is that because there's no equivalent service or it just wasn't necessary for your use case?

We could do the same (probably with BigQuery), but we just haven't needed it so far.

nmerket (approved):

PR #422, opened 2023-12-04
Move more shared code into DockerBatchBase

nweires:
## Pull Request Description

Moves two more sections of code into the parent class to be shared by the AWS and GCP implementations, and adds tests for them. Both are part of the tasks on the cloud that run the building simulations.

- `get_epws_to_download` determines which weather files are needed by a particular batch of simulations.
- `run_simulations` runs Openstudio, writes the results to the output bucket, and cleans up any intermediate files.

This is purely refactoring and does not change any behavior.

## Checklist

Not all may apply

- [x] Code changes (must work)
- [x] Tests exercising your feature/bug fix (check coverage report on Checks -> BuildStockBatch Tests -> Artifacts)
- [x] Coverage has increased or at least not decreased. Update `minimum_coverage` in `.github/workflows/coverage.yml` as necessary.
- [x] All other unit and integration tests passing
- [ ] ~~Update validation for project config yaml file changes~~
- [ ] ~~Update existing documentation~~
- [ ] ~~Run a small batch run on Kestrel/Eagle to make sure it all works if you made changes that will affect Kestrel/Eagle~~
- [x] Add to the changelog_dev.rst file and propose migration text in the pull request

PR #424, opened 2023-12-11
Two small fixes

nweires:
Fixes two small issues from #422 

## Pull Request Description

- Removes a stray comma
- Pass bucket/prefix as a single param (changed this in the test, but not aws.py)

## Checklist

Not all may apply

- [ ] ~~Add to the changelog_dev.rst file and propose migration text in the pull request~~

### DIFF
--- a/buildstockbatch/aws/aws.py
+++ b/buildstockbatch/aws/aws.py
@@ -1723,7 +1723,7 @@ class AwsBatch(DockerBatchBase):
                     logger.debug("Extracting {}".format(epw_filename))
                     f_out.write(gzip.decompress(f_gz.getvalue()))
 
-        cls.run_simulations(cfg, jobs_d, job_id, sim_dir, S3FileSystem(), bucket, prefix)
+        cls.run_simulations(cfg, jobs_d, job_id, sim_dir, S3FileSystem(), f"{bucket}/{prefix}")
 
 
 @log_error_details()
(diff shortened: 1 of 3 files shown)


buildstockbatch/aws/aws.py
```diff
@@ -1723,7 +1723,7 @@ def run_job(cls, job_id, bucket, prefix, job_name, region):
                     logger.debug("Extracting {}".format(epw_filename))
                     f_out.write(gzip.decompress(f_gz.getvalue()))
 
-        cls.run_simulations(cfg, jobs_d, job_id, sim_dir, S3FileSystem(), bucket, prefix)
+        cls.run_simulations(cfg, jobs_d, job_id, sim_dir, S3FileSystem(), f"{bucket}/{prefix}")
```
nmerket, line 1726: I see that in the `run_simulations` function definition there's just an `output_path` argument. Does that mean this wasn't already working?

nmerket (commented):
One question.

buildstockbatch/aws/aws.py
nweires: I changed the function params in response to [this comment](https://github.com/NREL/buildstockbatch/pull/422#discussion_r1414573255), but missed updating this line.

Meanwhile, I'm also looking into why the test didn't catch this.

nweires: Added some comments to test_docker_base.py to at least explain why the test there won't catch issues like the one in docker_base.py.

We should try to get that test to test things more thoroughly, but that's more involved so I'll save it for a separate PR.

nmerket (approved):

PR #426, opened 2024-01-02
handle empty data_point_out.json

rajeee:
## Pull Request Description

Sometimes, data_point_out.json can be present, but empty. This leads to obscure error downstream (see screenshot). This PR fixes the issue by treating empty json as not present. 

MicrosoftTeams-image (2)

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/postprocessing.py
+++ b/buildstockbatch/postprocessing.py
@@ -42,6 +42,8 @@ def read_data_point_out_json(fs, reporting_measures, filename):
     try:
         with fs.open(filename, "r") as f:
             d = json.load(f)
+        if not d:
+            return None
     except (FileNotFoundError, json.JSONDecodeError):
         return None
     else:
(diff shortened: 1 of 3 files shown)


buildstockbatch/postprocessing.py
```diff
@@ -42,6 +42,8 @@ def read_data_point_out_json(fs, reporting_measures, filename):
     try:
         with fs.open(filename, "r") as f:
             d = json.load(f)
+        if not d:
+            return None
     except (FileNotFoundError, json.JSONDecodeError):
         return None
```
nmerket, line 48: I'm kind of curious what kind of file would run into this block. We're already handling the file being missing and it not being valid json, which is the error hit when the file is just empty. Based on messing around a little with it, I think the file would have to be just

```json
null
```

I wonder where we're creating files like that?

rajeee: I ran into a case where it was just "{}" when ResStock fails in some specific step.

nmerket (approved):

PR #431, opened 2024-02-19
Checking if directory is deleted before copying

nmerket:
## Pull Request Description

Checks to see if the directory on local scratch is empty before copying to it.

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/hpc.py
+++ b/buildstockbatch/hpc.py
@@ -111,10 +111,15 @@ class SlurmBatch(BuildStockBatchBase):
     def clear_and_copy_dir(src, dst):
         if os.path.exists(dst):
             shutil.rmtree(dst, ignore_errors=True)
-        while True:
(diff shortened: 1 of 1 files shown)


buildstockbatch/hpc.py
```diff
@@ -111,7 +111,11 @@ def results_dir(self):
     def clear_and_copy_dir(src, dst):
         if os.path.exists(dst):
             shutil.rmtree(dst, ignore_errors=True)
-        shutil.copytree(src, dst)
+        while True:
```
nmerket: Maybe this should have a limit of the number of times it can run and error out so it doesn't hang forever if the file doesn't ever delete and it's not just a weird race condition. I'll fix that.

nmerket:
This didn't address the issue they were having as it wasn't a race condition. The actual fix is #438.

PR #435, opened 2024-02-22
Add function to log a summary of simulation statuses

nweires:
## Pull Request Description

Add a function to help log a summary of the simulations that ran during a job.

We're using this in the GCP implementation ([here](https://github.com/rewiringamerica/buildstockbatch/blob/65c87039594c4d878d8996518d0de07f433a4c60/buildstockbatch/gcp/gcp.py#L917)), but it should work with other systems too.

Example output:
```
INFO:2023-12-15 16:30:59:buildstockbatch.cloud.docker_base:Final status of all simulations:
Baseline     Success: 8        Fail: 0        Invalid: 0        
Upgrade 01   Success: 8        Fail: 0        Invalid: 0        
Upgrade 02   Success: 8        Fail: 0        Invalid: 0        
Upgrade 03   Success: 8        Fail: 0        Invalid: 0        
Upgrade 04   Success: 2        Fail: 0        Invalid: 6        
Upgrade 05   Success: 8        Fail: 0        Invalid: 0        
Upgrade 06   Success: 0        Fail: 0        Invalid: 8        
Upgrade 07   Success: 8        Fail: 0        Invalid: 0        
Upgrade 08   Success: 8        Fail: 0        Invalid: 0        
Upgrade 09   Success: 8        Fail: 0        Invalid: 0        
Upgrade 10   Success: 8        Fail: 0        Invalid: 0        
Upgrade 11   Success: 8        Fail: 0        Invalid: 0        
Upgrade 12   Success: 8        Fail: 0        Invalid: 0        

Total        Success: 90       Fail: 0        Invalid: 14       

No results found for Upgrade 13
No results found for Upgrade 14
```

## Checklist

Not all may apply


buildstockbatch/cloud/docker_base.py
```diff
@@ -461,3 +465,53 @@ def run_simulations(cls, cfg, job_id, jobs_d, sim_dir, fs, output_path):
                 shutil.rmtree(item)
             elif os.path.isfile(item):
                 os.remove(item)
+
+    def log_summary(self):
+        """
+        Log a summary of how many simulations succeeded, failed, or ended with other statuses.
+        Uses the `completed_status` column of the files in results_csvs/results_*.csv.gz.
+        """
```
nmerket: I like the premise of this. When does it get called? Also, is there any way this could be in base.py so it could be available to the other batch classes?

buildstockbatch/cloud/docker_base.py
```diff
@@ -461,3 +465,53 @@ def run_simulations(cls, cfg, job_id, jobs_d, sim_dir, fs, output_path):
...
+            results_files = fs.ls(results_csv_dir)
+        except FileNotFoundError:
+            logger.info(f"No results CSV files found at {results_csv_dir}")
+            return
+
+        for result in results_files:
+            upgrade_id = result.split(".")[0][-2:]
+            with fs.open(result) as f:
+                with gzip.open(f) as gf:
+                    df = pd.read_csv(gf, usecols=["completed_status"])
```
nmerket: The parquet files have the same data and are very fast and efficient to read, especially if you're only looking at a few columns.

nmerket (commented):
This is a great idea. I think we'd like this for all our runs.

buildstockbatch/cloud/docker_base.py
nweires: Good point - updated.

buildstockbatch/cloud/docker_base.py
nweires: In the GCP branch, we're calling it after post-processing finishes ([here](https://github.com/rewiringamerica/buildstockbatch/blob/65c87039594c4d878d8996518d0de07f433a4c60/buildstockbatch/gcp/gcp.py#L917)). (So this happens on your local machine, not on GCP.)

Moving it up to the parent class should work if the subclasses all add `get_fs()` to return a file system class that can read from `self.results_dir`. That would be easy for local and AWS, but I'm not sure about the others. What do you think?

nmerket (approved):
This looks good. We can wire this up to the other implementations in another PR.

PR #438, opened 2024-02-29
Fixes temp directory and bigmem issues with kestrel runs

ChristopherCaradonna:
Fixes # .

## Pull Request Description

This fixes Kestrel run errors related the existing files in the common temp directory on local node storage. Eagle assumes there is always local storage, and sets the TMPDIR to /tmp/scratch/$USER. Kestrel always assumes there is no local disk, and sets TMPDIR to /tmp/scratch. Existing files from other users on /tmp/scratch cause buildstock batch to fail.

This fix ensures the temporary directory is set to /tmp/scratch/$USER, avoiding file conflicts.

## Checklist

Not all may apply

- [x] ~~Tests exercising your feature/bug fix (check coverage report on Checks -> BuildStockBatch Tests -> Artifacts)~~


asparke2 (approved):
The real solution may be to get HPC to change the file permissions on /tmp/scratch, but this works in the meantime. @rHorsey suggested doing this via ENV in the kestrel shell scripts because then it doesn't affect Eagle.

nmerket (commented):
Looks good to me. Add and entry to the changelog and I'll merge it when you're ready.

nmerket:
Are we feeling like this is ready to merge? I'd like to get this into a release so people can benefit from these bugfixes.

asparke2:
@nmerket yeah, it's good to go, we used it for the Com EUSS runs

PR #445, opened 2024-03-27
Add script to sample a single location

nweires:
## Pull Request Description

Add a script that runs the residential quota sampler for a single county+PUMA.

Using the downselect sampler to do the same thing works, but it gets very slow when you filter aggressively. For example, generating 100 buildings for a single rural county took more than an hour to run. With this script, it takes just a few minutes. A similar option for ComStock exists [here](https://github.com/NREL/ComStock/blob/61f4bcb4d6355d0819094c6cd4da0180bfdd6d13/sampling/tsv_resampling_wrapper.py#L49).

## Checklist


shorowit:
@nweires I might be misunderstanding something, but isn't this what the [Residential Quota Downselect Sampler](https://buildstockbatch.readthedocs.io/en/stable/samplers/residential_quota_downselect.html) is for?

nweires:
> @nweires I might be misunderstanding something, but isn't this what the [Residential Quota Downselect Sampler](https://buildstockbatch.readthedocs.io/en/stable/samplers/residential_quota_downselect.html) is for?

Ah, yes, I should add some context for why this is useful! You can do the same thing with the downselect sampler, but this is much faster, especially rural locations with very few building. I'll put some notes about that in the script and PR description.

PR #449, opened 2024-04-19
Change LOCAL_SCRATCH from /tmp/scratch/$USER to /tmp/scratch/$JOBID

yingli-NREL:
Due to the change in the node local scratch configuration, to write to local disk, we now need to use the $TMPDIR variable. https://nrel.github.io/HPC/Documentation/Systems/Kestrel/kestrel_release_notes/


buildstockbatch/kestrel.sh
```diff
@@ -15,7 +15,7 @@ source "$MY_PYTHON_ENV/bin/activate"
 # Default LOCAL_SCRATCH = /tmp/scratch
 # Setting to user-specific dir to avoid
 # issues with deleting previous buildstock run debris
-export LOCAL_SCRATCH=/tmp/scratch/$USER
+export LOCAL_SCRATCH=/tmp/scratch/$SLURM_JOB_ID
```
rajeee: I think you should use $TMPDIR variable as suggested:
`export LOCAL_SCRATCH=$TMPDIR`

yingli-NREL: changed it to `export LOCAL_SCRATCH=$TMPDIR`

rajeee (approved):
This is following the directions here:
https://nrel.github.io/HPC/Documentation/Systems/Kestrel/filesystems/#node-file-system

PR #383, opened 2023-08-01
ResStock-HPXML: resilience and bill arguments

joseph-robertson:
Closes #381.

Support the following ReportSimulationOutput changes:
* Arguments `include_annual_resilience` and `include_timeseries_resilience` are now available. We will fix (i.e., user has no option to override) requesting annual resilience outputs. Requesting timeseries resilience outputs is optional.

Support the following ReportUtilityBills changes:
* Arguments `include_annual_bills`, `include_monthly_bills`, and `monthly_timestamp_convention` are now available. We will fix (i.e., user has no option to override) requesting annual utility bill outputs and _not_ requesting monthly utility bill outputs.

## Checklist

- [x] Code changes (must work)
- [x] Tests exercising your feature/bug fix (check coverage report on Checks -> BuildStockBatch Tests -> Artifacts)
- [x] Coverage has increased or at least not decreased. Update `minimum_coverage` in `.github/workflows/ci.yml` as necessary.
- [x] All other unit and integration tests passing
- [x] Update validation for project config yaml file changes
- [x] Update existing documentation
- [x] ~~Run a small batch run on Eagle to make sure it all works if you made changes that will affect Eagle~~
- [x] Add to the changelog_dev.rst file and propose migration text in the pull request

PR #451, opened 2024-05-20
ResStock-HPXML: optional utility bill reporting arguments

joseph-robertson:
## Pull Request Description

User now has the ability to optionally set utility bill reporting arguments:
- `include_annual_bills` (default: true)
- `include_monthly_bills` (default: false)

These will live under the `simulation_output_report` section of the yml file.

Argument `include_annual_bills=True` will generate `results_bills.csv` as well as register annual bills with the OpenStudio runner. Argument `include_monthly_bills` will generate `results_bills_monthly.csv` as well as register monthly bills with the OpenStudio runner.

(Related to https://github.com/NREL/resstock/pull/1246 and https://github.com/NREL/buildstockbatch/pull/383.)

## Checklist

Not all may apply

- [ ] ~Run a small batch run on Kestrel/Eagle to make sure it all works if you made changes that will affect Kestrel/Eagle~

### DIFF
--- a/buildstockbatch/workflow_generator/residential_hpxml.py
+++ b/buildstockbatch/workflow_generator/residential_hpxml.py
@@ -118,8 +118,6 @@ class ResidentialHpxmlWorkflowGenerator(WorkflowGeneratorBase):
             output_variables: list(include('output-var-spec'), required=False)
             include_annual_bills: bool(required=False)
             include_monthly_bills: bool(required=False)
-            register_annual_bills: bool(required=False)
-            register_monthly_bills: bool(required=False)
         output-var-spec:
             name: str(required=True)
         measure-spec:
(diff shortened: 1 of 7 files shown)


shorowit:
Does it really make sense to expose the new YAML inputs to a user? Why not just set both `include_foo` and `register_foo` to the same value based on the existing YAML input? For ResStock, doesn't seem like there's any harm to registering values unnecessarily...

buildstockbatch/workflow_generator/residential_hpxml.py
```diff
@@ -430,9 +434,27 @@ def create_osw(self, sim_id, building_id, upgrade_idx):
 
         if "include_annual_bills" in util_bills_rep_args_avail:
             util_bills_rep_args["include_annual_bills"] = True
```
joseph-robertson, line 435: Maybe we should now default this to False? buildstockbatch only ever got the runner-registered values -- so since we now have `register_annual_bills`, we may as well not write the `results_bills.csv` file, right?

joseph-robertson:
> Does it really make sense to expose the new YAML inputs to a user? Why not just set both `include_foo` and `register_foo` to the same value based on the existing YAML input? For ResStock, doesn't seem like there's any harm to registering values unnecessarily...

What's the existing YAML input? Currently these are hardset: `include_annual_bills=True` and `include_monthly_bills=False`.

If we hardset `register_monthly_bills=True`, couldn't your results.csv be crazy huge depending on how many bill scenarios are defined?

shorowit:
Oh sorry, I didn't realize there wasn't an input before. If you are adding inputs, I guess I'm advocating for two inputs instead of four. Only the `include_foo` arguments affect the results CSVs, right? The `register_foo` arguments only affect the results.json.

At the end of the day, the `register_foo` arguments are what ResStock needs to output utility bill results, right? I foresee a situation where a user sets `include_foo=true` and `register_foo=false`, runs millions of simulations, gets no utility bill results, and is not happy.

buildstockbatch/workflow_generator/residential_hpxml.py
shorowit: I think it's useful to write for inspection by a user. Much easier to read than the results.json.

joseph-robertson:
Hmm I do see your point. But on the flip side, seems like it'd be nice to save space by avoiding writing results_bills.csv and results_monthly_bills.csv.

afontani:
I want to point out that the actual way utility bills are reported is still mostly billing months. A bill dated March 1st is for February consumption. With the adoption of AMI metering, the bills are becoming much closer to calendar months. But utilities are still mixed in how they report bills. Some use the AMI timeseries aggregates for the month, others use the billing statements. Just a comment.

joseph-robertson:
I tested that this branch works [here](https://github.com/NREL/resstock/pull/1246) in resstock.

PR #466, opened 2024-08-14
Add --continue_upload flag

rajeee:
## Pull Request Description

Add --continue_upload flag so that if files are partially uploaded to s3, one can continue from that.

Verified operation by deleting some files from s3 and letting BSQ initialization flag it.
image

Then I ran --continue_upload to get the missing files uploaded. 
image

The error is then resolved.
image

--upload_only will still error out and suggest using --continue_upload.

image

## Checklist

Not all may apply


buildstockbatch/local.py
```diff
@@ -393,7 +393,13 @@ def main():
...
     group.add_argument(
         "--uploadonly",
-        help="Only upload to S3, useful when postprocessing is already done. Ignores the " "upload flag in yaml",
+        help="Only upload to S3, useful when postprocessing is already done. Ignores the upload flag in yaml."
+        " Errors out if files already exists in s3",
+        action="store_true",
+    )
+    group.add_argument(
+        "--continue_upload",
+        help="Continue uploading to S3, useful when previous upload was interrupted.",
```
nmerket, line 402: Why have two flags instead of wrapping this into the `--upload_only_` flag?

buildstockbatch/postprocessing.py
```diff
@@ -637,10 +637,14 @@ def upload_results(aws_conf, output_dir, results_dir, buildstock_csv_filename):
 
     s3 = boto3.resource("s3")
     bucket = s3.Bucket(s3_bucket)
-    n_existing_files = len(list(bucket.objects.filter(Prefix=s3_prefix_output)))
-    if n_existing_files > 0:
-        logger.error(f"There are already {n_existing_files} files in the s3 folder {s3_bucket}/{s3_prefix_output}.")
-        raise FileExistsError(f"s3://{s3_bucket}/{s3_prefix_output}")
+    existing_files = {f.key.removeprefix(s3_prefix_output) for f in bucket.objects.filter(Prefix=s3_prefix_output)}
```
nmerket, line 640: What happens when I file is partially uploaded and gets interrupted? Will it show up here even if it's not done?

rajeee: File uploads are atomic in s3. From AWS docs:

> Updates to a single key are atomic. For example, if you make a PUT request to an existing key from one thread and perform a GET request on the same key from a second thread concurrently, you will get either the old data or the new data, but never partial or corrupt data.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html#ConsistencyModel

buildstockbatch/local.py
rajeee: I added a new flag to preserve the existing behavior of --upload_only, which is to error out if there are existing files in s3.

buildstockbatch/hpc.py
```diff
@@ -563,7 +563,7 @@ def queue_jobs(self, array_ids=None, hipri=False):
         job_id = m.group(1)
         return [job_id]
 
-    def queue_post_processing(self, after_jobids=[], upload_only=False, hipri=False):
+    def queue_post_processing(self, after_jobids=[], upload_only=False, hipri=False, continue_upload=False):
```
wenyikuang, line 566: It's better don't set `[]` as default  value in python since it's mutable; https://github.com/satwikkansal/wtfpython?tab=readme-ov-file#-beware-of-default-mutable-arguments

wenyikuang (approved):

buildstockbatch/hpc.py
rajeee: Since our workflow calls queue_post_processing no more than once, it shouldn't be an issue. Regardless, agreed that it's a good practice to avoid mutable defaults.

PR #472, opened 2024-10-17
Switch to python 3.11

rajeee:
## Pull Request Description

Black stopped supporting python 3.8: https://github.com/psf/black/pull/4452. We can do the same as well, and make BSB use python 3.11 with support for 3.12. This allows us to use features introduced in later python version such as the [union operator on dict](https://peps.python.org/pep-0584/) and [fine-grained error location](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep657). 
## Checklist

Not all may apply


nmerket (commented):
Alternatively we could drop support for python 3.8 altogether and add support for 3.12. I'm able to run on 3.12, but it gives some warnings.

rajeee:
> Alternatively we could drop support for python 3.8 altogether and add support for 3.12. I'm able to run on 3.12, but it gives some warnings.

I considered doing that but remembered your reservation about dropping 3.8 because some part of BSB AWS needing to run inside the openstudio container.

nmerket (commented):
Any reason why we're dropping support for 3.9 and 3.10? It's officially supported until October 2025 and 2026, respectively.

Should also have a changelog entry.

nmerket (approved):

nmerket (approved):

Dockerfile
```diff
@@ -1,14 +1,17 @@
...
 FROM --platform=linux/amd64 nrel/openstudio:$OS_VER as buildstockbatch
 ARG CLOUD_PLATFORM=aws
-
-RUN sudo apt update && sudo apt install -y python3-pip
-RUN sudo -H pip install --upgrade pip
+ENV DEBIAN_FRONTEND=noninteractive
 COPY . /buildstock-batch/
-RUN python3 -m pip install "/buildstock-batch[${CLOUD_PLATFORM}]"
+
+RUN echo "hello"
```
nmerket:
```suggestion
```

Dockerfile
```diff
@@ -1,14 +1,17 @@
...
+RUN curl -k -LsSf https://astral.sh/uv/install.sh | sh
+ENV PATH="/root/.cargo/bin:$PATH"
+RUN uv python install 3.11
+RUN uv venv --python 3.11 && uv pip install "/buildstock-batch[${CLOUD_PLATFORM}]"
 
 # Base plus custom gems
-FROM buildstockbatch as buildstockbatch-custom-gems
+FROM --platform=linux/amd64 buildstockbatch as buildstockbatch-custom-gems
```
nmerket, line 15: Ah, was it trying to rebuild that base image on aarch64? Ugh, so annoying.

docs/changelog/changelog_dev.rst
```diff
@@ -117,3 +117,9 @@ Development Changelog
         Updates the bind mount for apptainer to include the whole resources
         directory instead of just the hpxml-measures directory. Makes it work
         with newer versions of ResStock.
+
+    .. change::
+        :tags: general
+        :pullreq: 472
+
+        Remove support for Python 3.8, 3.9, and 3.10.
```
nmerket:
```suggestion
        Remove support for Python 3.8, 3.9, and 3.10. Add support for Python 3.12.
```

setup.py
```diff
@@ -39,7 +39,7 @@
     long_description_content_type="text/markdown",
     url=metadata["__url__"],
     packages=setuptools.find_packages(),
-    python_requires=">=3.8",
+    python_requires=">=3.11",
```
nmerket: Does it work with Python 3.13 yet? Should we make that explicit here?

rajeee:
```suggestion
    python_requires=">=3.11, <3.13",
```

rajeee: It doesn't. I added that check.

Dockerfile
rajeee: Yeah, I was getting warnings in my machine.

Dockerfile
```diff
@@ -1,14 +1,16 @@
...
 FROM --platform=linux/amd64 nrel/openstudio:$OS_VER as buildstockbatch
 ARG CLOUD_PLATFORM=aws
-
-RUN sudo apt update && sudo apt install -y python3-pip
-RUN sudo -H pip install --upgrade pip
+ENV DEBIAN_FRONTEND=noninteractive
 COPY . /buildstock-batch/
-RUN python3 -m pip install "/buildstock-batch[${CLOUD_PLATFORM}]"
+
+RUN curl -k -LsSf https://astral.sh/uv/install.sh | sh
```
rajeee:
```suggestion
RUN curl -LsSf https://astral.sh/uv/install.sh | sh
```
Avoid insecure curl.

nmerket (approved):

PR #480, opened 2025-01-29
Update workflow generator to reference housing characteristics directly through project name

rajeee:
Fixes # .

## Pull Request Description
Pass project_directory to the BuildExistingModel and ApplyUpgrade measure in support of the changes in ResStock that get rids of the lib folder. Companion ResStock PR: https://github.com/NREL/resstock/pull/1342

## Checklist

Not all may apply


buildstockbatch/test/test_local.py
```diff
@@ -19,9 +19,8 @@
     "project_filename",
     [
         resstock_directory / "project_national" / "national_baseline.yml",
-        resstock_directory / "project_national" / "national_upgrades.yml",
         resstock_directory / "project_testing" / "testing_baseline.yml",
-        resstock_directory / "project_testing" / "testing_upgrades.yml",
+        # resstock_directory / "project_testing" / "testing_upgrades.yml",
```
joseph-robertson: Why not replace with `sdr_upgrades_tmy3.yml`?

rajeee: The sdr_upgrades_tmy3.yml has a different output data structure so the test won't pass. I need to work on that.

buildstockbatch/test/test_local.py
```diff
@@ -55,15 +55,22 @@ def test_resstock_local_batch(project_filename):
     simout_path = out_path / "simulation_output"
     assert (simout_path / "results_job0.json.gz").exists()
     assert (simout_path / "simulations_job0.tar.gz").exists()
-
+    upgrades2expected_bldgs = {
+        0: [1, 2],
+        1: [2],
+        2: [2],
+    }
```
joseph-robertson: Is this some kind of map to expect which datapoints/upgrades are "invalid" for the SDR upgrade yml file? If so, provide some comment here. If we were to bump n_datapoints for this yml in the future, would this test need to be updated?

rajeee: That's correct - basically, for the 1st two upgrades, which buildings are expected to apply. I can add a comment.

joseph-robertson (approved):
Add a changelog entry and I think this is probably good to go.

buildstockbatch/test/test_local.py
```diff
@@ -55,14 +55,18 @@ def test_resstock_local_batch(project_filename):
...
-    }
+
+    # Build upgrades2expected_bldgs map by scanning existing files
+    upgrades2expected_bldgs = {}  # key: upgrade_id, value: list of building_ids
+    for upgrade_dir in (simout_path / "timeseries").glob("up*"):
+        upgrade_id = int(upgrade_dir.name[2:])  # Extract ID from 'up##'
+        bldg_files = list(upgrade_dir.glob("bldg*.parquet"))
+        upgrades2expected_bldgs[upgrade_id] = [int(f.stem[4:]) for f in bldg_files]  # Extract ID from 'bldg#######'
+
+    # Verify that each upgrade has at least one building
```
joseph-robertson: As an alternative to making this test more robust, I would just make a note that this verification only applies to the first 2 upgrades in the SDR yml file. I.e., not _all_ upgrades currently in the SDR yml apply to at least 1 datapoint.

wenyikuang (approved):
Since the old generator is still in the `v2024_07_18` I think it's safe to merge.

buildstockbatch/test/test_local.py
```diff
@@ -33,9 +32,9 @@ def test_resstock_local_batch(project_filename):
     # Get the number of upgrades
     n_upgrades = len(batch.cfg.get("upgrades", []))
     # Limit the number of upgrades to 2 to reduce simulation time
```
wenyikuang: niche, change the `2` to `6` as you have changed.

PR #501, opened 2025-07-02
Add flag to replace existing files in s3

rajeee:
Used in https://github.com/NREL/resstock/pull/1439.

## Pull Request Description

Allows easy replacing of existing result in s3/Athena. Handy for CI to keep a s3 folder / Athena table up to date with latest result. 

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/postprocessing.py
+++ b/buildstockbatch/postprocessing.py
@@ -42,6 +42,7 @@ MAX_REPLACE_FILES = 9999  # maximum number of files to replace in s3 when using 
 # 1. It is inefficient
 # 2. It is easy to make mistakes and wipe out a significant run
 
+
 def read_data_point_out_json(fs, reporting_measures, filename):
     try:
         with fs.open(filename, "r") as f:
--- a/buildstockbatch/postprocessing.py
+++ b/buildstockbatch/postprocessing.py
@@ -669,7 +669,9 @@ def remove_intermediate_files(fs, results_dir, keep_individual_timeseries=False)
         fs.rm(ts_in_dir, recursive=True)
 
 
-def upload_results(aws_conf, output_dir, results_dir, buildstock_csv_filename, continue_upload=False):
+def upload_results(
+    aws_conf, output_dir, results_dir, buildstock_csv_filename, continue_upload=False, replace_existing=False
+):
     logger.info("Uploading the parquet files to s3")
 
     output_folder_name = Path(output_dir).name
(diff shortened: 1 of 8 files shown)


buildstockbatch/postprocessing.py
```diff
@@ -698,8 +700,12 @@ def upload_results(aws_conf, output_dir, results_dir, buildstock_csv_filename, c
 
     if len(existing_files) > 0:
         logger.info(f"There are already {len(existing_files)} files in the s3 folder {s3_bucket}/{s3_prefix_output}.")
-        if not continue_upload:
-            raise FileExistsError("Either use --continue_upload or delete files from s3")
+        if not continue_upload and not replace_existing:
+            raise FileExistsError("Either use --continue_upload or --replace_existing or delete files from s3")
+        if replace_existing:
+            bucket.objects.filter(Prefix=s3_prefix_output).delete()
```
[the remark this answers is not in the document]
rajeee: Addressed by adding MAX_REPLACE_FILES

buildstockbatch/test/test_postprocessing.py
```diff
@@ -197,3 +196,73 @@ def publish_upgrade_annual_results(failed_bldgs, base, upgrade, upgrade_num):
         upgrade_id = int(upgrade_folder.name.split("=")[1])
         expected_file = upgrade_folder / f"results_up{upgrade_id:02d}.parquet"
         assert expected_file.exists(), f"Expected parquet file missing for {upgrade_folder.name}"
+
+
+@pytest.mark.parametrize(
+    "scenario",
+    [
+        {"replace_existing": True, "continue_upload": False, "should_raise_error": False},
+        {"replace_existing": False, "continue_upload": False, "should_raise_error": True},
```
afontani, line 205: If I read this right, It seems like this should fail. You need to have either of these specified? Another way is to combine the arguments and default to replace_existing being True. Just a thought.

afontani (approved):
I think this will work okay. The tests are passing. I made a comment for consideration. Another thing that might be useful is a replacement log or a log file for upload_results.

buildstockbatch/test/test_postprocessing.py
rajeee: Correct. That's why should_raise_error is True for this combination and we verify that happens. It should normally never get to this point though since the arg_parser will complain about two options from mutually exclusive groups being provided. 
image

PR #512, opened 2025-08-07
Memory efficiency tweak

rajeee:
Fixes # .

## Pull Request Description

Simplifies the postprocessing by writing partitioned annual parquet files. This reduces the memory consumption during postprocessing because it is no longer necessary to read all the results_jobxx.json files to get the result for a particular upgrade. They can be simply read by reading all the parquet files in the simulation_ouput/annual/upxx/ folder. 

In addition, few other changes (probably should have been a separate PR ...)
1. Intermediate results are no longer removed. So, postprocessing can be done again on a run.
2. weathe_file_path can point to either a directory of weather files or a zip file. Previously, you could only point to zip file. Pointing to unzipped directory skips the unzip step and reduces the wait time before simulation is started.
3. Add --ultra-low-disk-no-timeseries flag to buildstock_local.  This was necessary to make BSB simulation sucessfully run with 37 buildings and 17 upgrades in: https://github.com/NREL/resstock/pull/1462

## Checklist

Not all may apply


buildstockbatch/base.py
```diff
@@ -107,11 +107,18 @@ def path_rel_to_projectfile(self, x):
 
     def _get_weather_files(self):
         if "weather_files_path" in self.cfg:
-            logger.debug("Copying weather files")
             weather_file_path = self.cfg["weather_files_path"]
-            with zipfile.ZipFile(weather_file_path, "r") as zf:
-                logger.debug("Extracting weather files to: {}".format(self.weather_dir))
-                zf.extractall(self.weather_dir)
+            if os.path.isdir(weather_file_path):
```
rajeee, line 112: weather file path now can be directory too (not just a zipfile). Will make buildstock_local much faster if using pre-unzipped directory.

asparke2: Sweet!

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         with:
           repository: NREL/resstock
           path: resstock
-          ref: develop
+          ref: better-minimal-buildstock
```
asparke2, line 31: Assume this gets changed back after resstock gets updated too.

asparke2 (approved):

buildstockbatch/utils.py
```diff
@@ -153,3 +167,12 @@ def get_annual_publishing_functions(stock_type):
         return publish_baseline_annual_results, publish_upgrade_annual_results
     else:
         raise ValueError(f"Stock type: {stock_type} currently does not support postprocessing transform")
+
+
+def get_data_dict_schema(stock_type, all_cols):
+    if stock_type == "residential":
+        from resstockpostproc import get_polars_schema_from_data_dictionary
```
asparke2: Will add ComStock version of this function in comstockpostprocessing.

asparke2 (approved):

.github/workflows/ci.yml
rajeee: Yeah, this PR relies on some changes on resstock currently only available on that branch. These two branches are mutually tied up together - one needs the other. So, once we are ready on both sides, we switch the branch to develop on both sides and merge both side in quick order.

PR #522, opened 2026-03-25
Fix pandas 3.x compatibility: cast building ID index to int after CSV read

billyhines:
Fixes # 521

## Pull Request Description

In pandas 3.x, `dtype=str` in `read_csv()` now strictly applies to the index when `index_col=0` is used, causing building IDs to be returned as strings instead of integers. This breaks the format string `"bldg{:07d}up{:02d}".format(building_id, ...)` in `base.py` with a `ValueError: Unknown format code 'd' for object of type 'str'`.

This fix explicitly casts the DataFrame index back to `int` immediately after reading the buildstock CSV in all three execution paths (local, HPC, and cloud/Docker) as well as in sampler/downselect.py. The fix is backward-compatible with pandas 2.x, where `.astype(int)` is harmless since the index was already integer.

Two previously failing tests (test_resstock_local_batch[national_baseline] and test_resstock_local_batch[testing_baseline]) now pass with this fix.

**Files changed:**
- `buildstockbatch/local.py`
- `buildstockbatch/hpc.py`
- `buildstockbatch/cloud/docker_base.py`
- `buildstockbatch/sampler/downselect.py`

## Checklist

Not all may apply

### DIFF
--- a/buildstockbatch/cloud/docker_base.py
+++ b/buildstockbatch/cloud/docker_base.py
@@ -431,6 +431,7 @@ class DockerBatchBase(BuildStockBatchBase):
 
         logger.debug("Validating sampled buildstock...")
         df = read_csv(buildstock_csv_filename, index_col=0, dtype=str)
+        df.index = df.index.astype(int)
         self.validate_buildstock_csv(self.project_filename, df)
         building_ids = df.index.tolist()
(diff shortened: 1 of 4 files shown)


shorowit:
Thanks @billyhines. I've pulled your commit into #523 to test with our CI.

